PR #17, opened 2016-07-28
readme - permissions fix

zjalexander:
If you try and use Notepad.exe  or Visual Studio to save the psm1 file, you will get an error about insufficient permissions even if you are logged in as Admin.
I worked around this by using ISE as Administrator

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Install-Module -Name PowerShellForGitHub
 1) Rename ApiTokensTemplate.psm1 to ApiTokens.psm1 and update value of $global:gitHubApiToken with GitHub token for your account
   * You can obtain it from https://github.com/settings/tokens). 
   * If you don't provide GitHub token, you can still use this module, but you will be limited to 60 queries per hour. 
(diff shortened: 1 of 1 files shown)


KarolKaczmarek:
@zjalexander 
Thanks for filing this.
Did you hit this issue when installing module from PowerShell Gallery via Install-Module (directly to C:\Program Files)? 

That would explain why you were required to provide administrator rights to rename the file. However, when I did that I was able to rename (save as) the file either via notepad.exe (elevated), PowerShell ISE or simply by right clicking and choosing "rename". 

When I tried to save it as "ApiTokens.psm1" using non-elevated notepad.exe it didn't work as expected and proposed to save it my user's folder instead.

zjalexander:
Discussed briefly with Karol:
1) yes, used Install-Module
2)  I didn't think through finding the notepad shortcut and right clicking to run as admin :) 
given that, I slightly changed the wording

KarolKaczmarek:
Thanks!

PR #25, opened 2018-01-11
Update README.md

aedahl:
Removed trailing 's'


aedahl:
On my end, it is showing that I have signed the CLA.

Any troubleshooting suggestions?

PR #31, opened 2018-07-13
new commandlet introduced: Get-GitHubForksForRepository, also fixed error with paging

chipitsine:
I added commandlet for enumerating forks (I wished to get all forks for very big repo, it is not available via web gui), when doing that I also noticed that paging was working in wrong way

### DIFF
--- a/GitHubAnalytics.psm1
+++ b/GitHubAnalytics.psm1
@@ -1211,3 +1211,66 @@ function Get-WeekDate
 
     return $beginningsOfWeeks
 }
+
+<#
+    .SYNOPSIS Obtain repository forks from it's url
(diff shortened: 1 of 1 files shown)


HowardWolosky:
Thanks so much for this submission.
As you can see, I've just taken over ownership for this module since it was abandoned, and I've been doing some general housecleaning to make it easier to consistently add new functionality.  Your submission, while great, was against the old structure of the module.

I've gone ahead and added the requested functionality (querying forks), along with adding related functionality (creating a new fork), along with the associated tests and documentation.

Please do consider contributing additional changes to the module in the future now that things are all in order!

Issue #33, opened 2018-07-25
TLS error in Get-GitHubIssueForRepository : Failed to execute query with exception


erichexter:
It looks like there is a issue with TLS with the latest version published to the PSGallery.

When running 
```
$issues=Get-GitHubIssueForRepository -repositor
yUrl @('https://github.com/QuarterSpot/../')
Getting issues for repository https://github.com/QuarterSpot/.../

```

I get the following error:

```
Get-GitHubIssueForRepository : Failed to execute query with exception: System.Net.WebException: The request was
aborted: Could not create SSL/TLS secure channel.
   at Microsoft.PowerShell.Commands.WebRequestPSCmdlet.GetResponse(WebRequest request)
   at Microsoft.PowerShell.Commands.WebRequestPSCmdlet.ProcessRecord()
HTTP status code:
At line:1 char:9
+ $issues=Get-GitHubIssueForRepository -repositoryUrl @('https://github ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException,Get-GitHubIssueForRepository

```

I found a workaround which is to set the default security protocol. I just executed this in the same session that the module was imported.

`[System.Net.ServicePointManager]::SecurityProtocol=[System.Net.SecurityProtocolType]::Tls12`

HowardWolosky:
Fixed in 0.2.0

PR #35, opened 2018-10-07
Adding TLS 1.2 support

zshehri:
Github [discontinued support for weak cryptographic standards](https://githubengineering.com/crypto-deprecation-notice/) last year. Therefore, TLS 1.2 should be enforced on all Rest API calls in the PowerShell modules.

This update also fixes [#33](https://github.com/PowerShell/PowerShellForGitHub/issues/33).

### DIFF
--- a/GitHubAnalytics.psm1
+++ b/GitHubAnalytics.psm1
@@ -90,6 +90,8 @@ function Get-GitHubIssueForRepository
         {
             try
             {
+                # Force TLS 1.2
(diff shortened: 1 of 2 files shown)


HowardWolosky:
Thanks so much for the submission.  I just took over administration of this module, and I'm currently refactoring the code to centralize the REST interaction in order to minimize duplicated code (and make it easier to add additional functions).  I will be fixing Issue #33 with that update, but I don't want to bring this change in as-is since so much of that code will be going away.

PR #49, opened 2018-11-14
Add Support for the Github Traffic APIs

joseartrivera:
Added support for all of the [Github Traffic APIs](https://developer.github.com/v3/repos/traffic/).

Also added GitHubTraffic.tests.ps1 for testing. These tests will only ensure that we can still talk to the Github APIs since there is not a reliable way to test the data that the end point would return on a new or existing repository.


GitHubTraffic.ps1
```diff
@@ -0,0 +1,344 @@
...
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/traffic/popular/referrers"
+        'Method' = 'Get'
+        'Description' =  "Get top 10 referrers for $RepositoryName"
+        'AcceptHeader' = 'application/vnd.github.symmetra-preview+json'
```
HowardWolosky: You don't need a beta `AcceptHeader` for any of these commands.

GitHubTraffic.ps1
```diff
@@ -0,0 +1,344 @@
...
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/traffic/popular/referrers"
+        'Method' = 'Get'
+        'Description' =  "Get top 10 referrers for $RepositoryName"
```
HowardWolosky: This description by default shows up when the command is running asynchronously.  To avoid the commandline going crazy, it's a good idea to keep them short....so.... `Getting referrers for $RepositoryName` would be good here, and a similarly short description for the others.

GitHubTraffic.ps1
```diff
@@ -0,0 +1,344 @@
...
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/traffic/views"
```
HowardWolosky: [This API](https://developer.github.com/v3/repos/traffic/#views) supports the `per` parameter.  Can you add support for that as well?

GitHubTraffic.ps1
```diff
@@ -0,0 +1,344 @@
...
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/traffic/clones"
```
HowardWolosky: [This API](https://developer.github.com/v3/repos/traffic/#clones) supports the `per` parameter.  Can you add support for that as well?

Tests/GitHubTrafficSupport.tests.ps1
```diff
@@ -0,0 +1,133 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
HowardWolosky: Looks like this file is an accident?  Seems duplicative of the other test file.

joseartrivera: Yes, this is now deleted

GitHubTraffic.ps1
```diff
@@ -0,0 +1,362 @@
...
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+        'Per' = $Per
+    }
+
+    $getParams = @(
+        "per=$Per"
+    )
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/traffic/views`?" +  ($getParams -join '&')
```
danbelcher-MSFT: Seems like `$getParams` is used elsewhere with multiple elements and we're just matching a style here.  It seems unnecessary and I would support just adding `"per=$Per"` to the rest of the UriFragment.

HowardWolosky: Totally reasonable feedback.  I'm preparing to publish the 0.4.0 update.  I can make that minor change.

PR #53, opened 2018-11-26
Add support for comment APIs

joseartrivera:
Added support for all of the [Github Comment APIs](https://developer.github.com/v3/issues/comments/).


GitHubComments.ps1
```diff
@@ -62,33 +62,37 @@ function Get-GitHubIssueComment
...
 
-    Write-InvocationLog -Invocation $MyInvocation
+    Write-InvocationLog
 
-    $elements = Resolve-RepositoryElements -BoundParameters $PSBoundParameters
+    $elements = Resolve-RepositoryElements
     $OwnerName = $elements.ownerName
     $RepositoryName = $elements.repositoryName
 
+    if ($Since -ne $null) {
```
danbelcher-MSFT: nit: style of the codebase is to put curlies on a newline

danbelcher-MSFT: PowerShell best-practice is to always compare against `$null` with `$null` on the left, that is, `$null -ne $Since`

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+        'Number' =  $IssueNumber
+        'Since' = $Since
+    }
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/issues/$IssueNumber/comments`?since=$Since"
+        'Description' =  "Getting comments for issue $IssueNumber in $RepositoryName"
+        'AccessToken' = $AccessToken
```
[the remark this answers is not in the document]
HowardWolosky: At the [bottom of the docs](https://developer.github.com/v3/issues/comments/#custom-media-types), it also talks about other media types that could be used.  You might want to play with the idea of having an optional parameter that lets users choose between the content type returned (raw, text, html, full, reactions).  You should see what the returned content looks like amongst those different values to see if that's worthwhile.

GitHubComments.ps1
```diff
@@ -62,33 +62,37 @@ function Get-GitHubIssueComment
...
+        $SinceFormattedTime = $Since.ToUniversalTime().ToString('o')
+    }
+
     $telemetryProperties = @{
         'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
         'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
-        'Number' =  $IssueNumber
-        'Since' = $Since
+        'IssueNumber' =  $IssueNumber
+        'Since' = $SinceFormattedTime
```
HowardWolosky: For things like this, I don't think it matters as much what this value was.  From a telemetry perspective, we'd more possibly be interested in _if_ they provided it at all.  You'll see [other instances](https://github.com/PowerShell/PowerShellForGitHub/blob/master/GitHubIssues.ps1#L179) where we just log whether or not the user provided a value for a parameter, as opposed to the value of the parameter itself.  It's more about being able to answer the question: "how are people using this function".  Tracking the OnwerName/RepositoryName allows us to identify the breadth of the usage across GitHub, while things like this allow us to track how an individual function is getting used.

GitHubComments.ps1
```diff
@@ -178,7 +179,8 @@ function Get-GitHubRepositoryComment
     $OwnerName = $elements.ownerName
     $RepositoryName = $elements.repositoryName
 
-    if ($Since -ne $null) {
+    if ($null -ne $Since) 
+    {
         $SinceFormattedTime = $Since.ToUniversalTime().ToString('o')
```
HowardWolosky: I still think you should align on how this is handled in [GitHubIssues](https://github.com/PowerShell/PowerShellForGitHub/blob/master/GitHubIssues.ps1#L233) by not sending the parameter at all if no value has been specified. The API doesn't provide guarantees for how it would treat that type of input, so we should be crisp in the data that we do/do not send.

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+        'AccessToken' = $AccessToken
+        'TelemetryEventName' = $MyInvocation.MyCommand.Name
+        'TelemetryProperties' = $telemetryProperties
+        'NoStatus' = (Resolve-ParameterWithDefaultConfigurationValue -BoundParameters $PSBoundParameters -Name NoStatus -ConfigValueName DefaultNoStatus)
+    }
+
+    return Invoke-GHRestMethod @params
+}
+
+function Edit-GitHubComment
```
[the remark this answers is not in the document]
joseartrivera: Set makes sense, updated it to be consistent with the rest of the project.

GitHubComments.ps1
joseartrivera: That makes sense, updated this to be a bool

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+        'Number' =  $IssueNumber
+        'Since' = $Since
+    }
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/issues/$IssueNumber/comments`?since=$Since"
```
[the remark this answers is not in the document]
joseartrivera: Rest APIs shouldn't fail with empty query params but I can add in checks to be consistent and clear.

joseartrivera:
@HowardWolosky I will tackle repository comments (https://developer.github.com/v3/repos/comments/) in a seperate PR

GitHubComments.ps1
```diff
@@ -96,6 +109,11 @@ function Get-GitHubComment
         [ValidateSet('asc', 'desc')]
         [string] $Direction,
 
+        [ValidateSet('raw', 'text', 'html', 'full', '')]
+        [string] $MediaType,
```
danbelcher-MSFT: If `raw` is the default, you can specify that as a default value for this parameter and remove the empty string case:
```powershell
[ValidateSet('raw', 'text', 'html', 'full')]
[string] $MediaType = 'raw'
```

GitHubComments.ps1
```diff
@@ -139,33 +61,58 @@ function Get-GitHubRepositoryComment
...
+        [Parameter(ParameterSetName='RepositoryUri')]
+        [Parameter(ParameterSetName='RepositoryElements')]
         [ValidateSet('asc', 'desc')]
         [string] $Direction,
 
-        [DateTime] $Since,
+        [ValidateSet('raw', 'text', 'html', 'full', '')]
+        [string] $MediaType,
+
+        [string] $MediaTypeVersion = 'VERSION',
```
danbelcher-MSFT: There are better ways to determine if the user specified a value or not.  I would recommend:
```powershell
[ValidateNotNullOrEmpty()]
[string] $MediaTypeVersion = ''
```
The validation specified _only_ happens if the caller uses this parameter and _does not_ happen for the default value you supplied. This means the caller is _forced_ to provide a valid, nonempty version string if they use this parameter and, in your code, a value of empty string means the caller did not supply a value
```powershell
Get-GitHubComment @args -MediaTypeVersion 'someVersion' # works
Get-GitHubComment @args -MediaTypeVersion '' # fails
Get-GitHubComment @args -MediaTypeVersion $null # fails
Get-GitHubComment @args # -MediaTypeVersion not specified, will be empty string inside the function's code.
```

joseartrivera: If this param is not supplied it should default to 'VERSION' since this is what is used in the docs when the latest version is requested. I am not too worried about validating if they pass null as invalid values will just default to the latest version being returned.

danbelcher-MSFT: Ok, that makes sense then.

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+        'IssueNumber' =  (Get-PiiSafeString -PlainText $IssueNumber)
+    }
+
+    $hashBody = @{
+        'body' = $Body
+    }
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/issues/$IssueNumber/comments"
+        'Body' = ($hashBody | ConvertTo-Json)
```
HowardWolosky: For consistency with recent changes, please make this: `(ConvertTo-Json -InputObject $hashBody)`

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+    .PARAMETER NoStatus
+        If this switch is specified, long-running commands will run on the main thread
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .EXAMPLE
+        Set-GitHubComment -OwnerName Powershell -RepositoryName PowerShellForGitHub -CommentID 1 -Body "Testing this API"
+
+        Set an existing comment in an issue for the PowerShell\PowerShellForGitHub project.
```
HowardWolosky: I'd rephrase this as "Update an existing...."  I think there's a different between how PowerShell users name/think of methods, and how we'd _describe_ those methods.

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+        [Parameter(Mandatory)]
+        [string] $CommentID,
+
+        [Parameter(Mandatory)]
+        [string] $Body,
+
+        [ValidateSet('raw', 'text', 'html', 'full')]
+        [string] $MediaType ='raw',
+
+        [string] $MediaTypeVersion = 'VERSION',
```
HowardWolosky: Again, i think we should drop this

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+        [Parameter(Mandatory, ParameterSetName='CommentUri')]
+        [string] $Uri,
+
+        [Parameter(Mandatory, ParameterSetName='CommentUri')]
+        [Parameter(Mandatory, ParameterSetName='CommentElements')]
+        [string] $CommentID,
+
+        [Parameter(Mandatory, ParameterSetName='IssueUri')]
+        [Parameter(Mandatory, ParameterSetName='IssueElements')]
+        [int] $IssueNumber,
```
HowardWolosky: The `GitHubIssues.ps1` functions just use `Issue` as the parameter name.  I'd stay consistent.

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+    }
+
+    return Invoke-GHRestMethod @params
+}
+
+function Set-GitHubComment
+{
+<#
+    .DESCRIPTION
+        Set an existing comment in an issue for the given repository
```
HowardWolosky: "Set an ..." -> "Update an ..."

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+
+    .PARAMETER Uri
+        Uri for the repository.
+        The OwnerName and RepositoryName will be extracted from here instead of needing to provide
+        them individually.
+
+    .PARAMETER CommentID
+        The comment ID of the comment to edit.
+
+    .PARAMETER body
```
HowardWolosky: `body` -> `Body`

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+
+    $hashBody = @{
+        'body' = $Body
+    }
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/issues/comments/$CommentID"
+        'Body' = ($hashBody | ConvertTo-Json)
+        'Method' = 'Patch'
+        'Description' =  "Set comment $CommentID for $RepositoryName"
```
HowardWolosky: "Set" -> "Update"

GitHubComments.ps1
```diff
@@ -0,0 +1,576 @@
...
+
+    .EXAMPLE
+        Remove-GitHubComment -OwnerName Powershell -RepositoryName PowerShellForGitHub -CommentID 1
+
+        Deletes a Github comment from the PowerShell\PowerShellForGitHub project.
+#>
+    [CmdletBinding(
+        SupportsShouldProcess,
+        DefaultParametersetName='Elements')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "", Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
```
HowardWolosky: You should probably add a [`Delete-GitHubComment` alias](https://github.com/PowerShell/PowerShellForGitHub/blob/master/GitHubRepositories.ps1#L224)  to this one, to be consistent with how we treat the other `Remove-*` methods (you'll also need to add it to the .psd1).

GitHubComments.ps1
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+$script:MediaTypeVersion = 'v3'
```
HowardWolosky: Move this to `GitHubCore.ps1` (it will still be accessible from here)

GitHubComments.ps1
```diff
@@ -0,0 +1,545 @@
...
+#>
+    [CmdletBinding()]
+    param(
+        [ValidateSet('raw', 'text', 'html', 'full')]
+        [string] $MediaType ='raw'
+    )
+
+    $acceptHeaders = @('application/vnd.github.squirrel-girl-preview')
+
+    if ('raw' -eq $MediaType)
```
HowardWolosky: I think you can simplify this whole block to just this (given that you know it will always have a value, and your valid types are equivalent to those used in the header):

```powershell
 $acceptHeaders += "application/vnd.github.$script:MediaTypeVersion.$MediaType+json"
```

GitHubComments.ps1
danbelcher-MSFT: Prefer `Set-Variable -Scope Script ...` so that you can use other options like `ReadOnly`

danbelcher-MSFT: If your script starts with multiple variable declarations, you can use a hashtable to make the process more efficient:
```powershell
# Define script-scope, readonly variables
@{
   # myVar1 is useful for ...
   myVar1 = 'hello'

   # version is used for the API..
    requestVersion = 'v3'

}.GetEnumerator() | ForEach-Object {
    Set-Variable -Scope Script -Option ReadOnly -Name $_.Key -Value $_.Value
}
```

GitHubComments.ps1
joseartrivera: Good idea

PR #59, opened 2018-12-04
Add support for issue labels

joseartrivera:
Added functionality for the [issue label APIs](https://developer.github.com/v3/issues/labels/)

### DIFF
--- a/GitHubLabels.ps1
+++ b/GitHubLabels.ps1
@@ -663,10 +663,10 @@ function Add-GitHubIssueLabel
         DefaultParametersetName='Elements')]
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "", Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
     param(
-        [Parameter(ParameterSetName='Elements')]
+        [Parameter(Mandatory, ParameterSetName='Elements')]
         [string] $OwnerName,
 
-        [Parameter(ParameterSetName='Elements')]
+        [Parameter(Mandatory, ParameterSetName='Elements')]
         [string] $RepositoryName,
 
         [Parameter(
--- a/GitHubLabels.ps1
+++ b/GitHubLabels.ps1
@@ -61,9 +61,9 @@ function Get-GitHubLabel
         DefaultParametersetName='Elements')]
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "", Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
     param(
-        [Parameter(Mandatory,ParameterSetName='Elements')]
-        [Parameter(Mandatory,ParameterSetName='NameElements')]
-        [Parameter(Mandatory,ParameterSetName='IssueElements')]
+        [Parameter(Mandatory, ParameterSetName='Elements')]
+        [Parameter(Mandatory, ParameterSetName='NameElements')]
+        [Parameter(Mandatory, ParameterSetName='IssueElements')]
         [Parameter(Mandatory, ParameterSetName='MilestoneElements')]
         [string] $OwnerName,
 
@@ -681,8 +681,6 @@ function Add-GitHubIssueLabel
         [Alias('LabelName')]
         [string[]] $Name,
 
-        [switch] $Replace,
-
         [string] $AccessToken,
 
         [switch] $NoStatus
@@ -697,7 +695,7 @@ function Add-GitHubIssueLabel
     $telemetryProperties = @{
         'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
         'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
-        'labelCount' = $Name.Count
+        'LabelCount' = $Name.Count
     }
 
     $hashBody = @{
@@ -784,8 +782,6 @@ function Set-GitHubIssueLabel
         [Alias('LabelName')]
         [string[]] $Name,
 
-        [switch] $Replace,
-
         [string] $AccessToken,
 
         [switch] $NoStatus
@@ -800,7 +796,7 @@ function Set-GitHubIssueLabel
     $telemetryProperties = @{
         'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
         'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
-        'labelCount' = $Name.Count
+        'LabelCount' = $Name.Count
     }
 
     $hashBody = @{
@@ -886,8 +882,6 @@ function Remove-GitHubIssueLabel
         [ValidateNotNullOrEmpty()]
         [Alias('LabelName')]
         [string] $Name,
-
-        [switch] $RemoveAll,
 
         [string] $AccessToken,
 
--- a/GitHubLabels.ps1
+++ b/GitHubLabels.ps1
@@ -29,6 +29,12 @@ function Get-GitHubLabel
         Name of the specific label to be retieved.  If not supplied, all labels will be retrieved.
         Emoji and codes are supported.  For more information, see here: https://www.webpagefx.com/tools/emoji-cheat-sheet/
 
+    .PARAMETER Issue
+        If provided, will return all of the labels for this particular issue.
+
+    .PARAMETER Milestone
+        If provided, will return all of the labels for this particular milestone.
+
     .PARAMETER AccessToken
         If provided, this will be used as the AccessToken for authentication with the
         REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
@@ -69,6 +75,10 @@ function Get-GitHubLabel
         [Alias('LabelName')]
         [string] $Name,
 
+        [int] $Issue,
+
+        [int] $Milestone,
+
         [string] $AccessToken,
 
         [switch] $NoStatus
@@ -85,9 +95,33 @@ function Get-GitHubLabel
         'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
     }
 
+    if ($PSBoundParameters.ContainsKey('Issue'))
+    {
+        $uriFragment = "/repos/$OwnerName/$RepositoryName/issues/$Issue/labels"
+        $description = "Getting labels for Issue $Issue in $RepositoryName"
+    }
+    elseif ($PSBoundParameters.ContainsKey('Milestone'))
+    {
+        $uriFragment = "/repos/$OwnerName/$RepositoryName/milestones/$Issue/labels"
+        $description = "Getting labels for Milestone $Milestone in $RepositoryName"
+    }
+    else
+    {
+        $uriFragment = "repos/$OwnerName/$RepositoryName/labels/$Name"
+
+        if ($PSBoundParameters.ContainsKey('Name'))
+        {
+            $description =  "Getting label $Name for $RepositoryName"
+        }
+        else
+        {
+            $description = "Getting labels for $RepositoryName"
+        }
+    }
+
     $params = @{
-        'UriFragment' = "repos/$OwnerName/$RepositoryName/labels/$Name"
-        'Description' =  "Getting all labels for $RepositoryName"
+        'UriFragment' = $uriFragment
+        'Description' =  $description
         'AcceptHeader' = 'application/vnd.github.symmetra-preview+json'
         'AccessToken' = $AccessToken
         'TelemetryEventName' = $MyInvocation.MyCommand.Name
@@ -251,6 +285,9 @@ function Remove-GitHubLabel
         Name of the label to be deleted.
         Emoji and codes are supported.  For more information, see here: https://www.webpagefx.com/tools/emoji-cheat-sheet/
 
+    .PARAMETER Issue
+        Issue number to delete the label from. If not provided the label will be deleted from the entire repository.
+
     .PARAMETER AccessToken
         If provided, this will be used as the AccessToken for authentication with the
         REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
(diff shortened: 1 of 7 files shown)


GitHubLabels.ps1
```diff
@@ -565,6 +628,121 @@ function Set-GitHubLabel
     }
 }
 
+function Add-GitHubLabel
```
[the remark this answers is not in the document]
danbelcher-MSFT: What's the common scenario?  Let's say I want to add a label to something.  Does that require a call to New-GithubLabel and Add-GithubLabel?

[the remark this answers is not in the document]
HowardWolosky: Adding new labels is uncommon.  Adding labels to issues is common and frequent.  I think that separating this functionality out between two separate methods makes sense.  It does raise the question on why we'd then have a merged `Remove-GitHubLabel`.  From a naming perspective, it makes sense in both scenarios, but it is I guess odd that we'd have a converged 'Remove' but a separated 'Add'/'New'

GitHubLabels.ps1
```diff
@@ -565,6 +628,121 @@ function Set-GitHubLabel
...
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+        'labelCount' = $LabelName.Count
+    }
+
+    $hashBody = @{
+        'labels' = $LabelName
+    }
+
+    if ($Replace)
```
danbelcher-MSFT:
```powershell
$method = if ($Replace) { 'Put' }
          else { 'Post' }
```

Tests/GitHubLabels.tests.ps1
```diff
@@ -271,6 +271,61 @@ if ($script:accessTokenConfigured)
 
         $null = Remove-GitHubRepository -OwnerName $script:ownerName -RepositoryName $repositoryName
     }
+
+    Describe 'Adding and removing labels to an issue'{
```
danbelcher-MSFT: It doesn't feel like `Describe`, `Context`, and `It` are being used correctly here.  My understanding is that `Describe` is intended to be scoped for a single function.  So `Describe 'Add-GitHubLabel'`...  `It 'does this thing...'` and `It 'does this other thing...'`.  Optionally you can add in `Context` for when the behavior is dependent on other state, e.g. `Context: 'Label already exists'`.  You can _use_ these constructs as it is here, but it doesn't seem to match the intent of how they should be used.

GitHubLabels.ps1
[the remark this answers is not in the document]
joseartrivera: @HowardWolosky I agree, would it make more sense to just have two separate functions called Add-GitHubIssueLabel and Remove-GitHubIssueLabel? Does it also make more sense to move these in with the rest of the Issues code?

GitHubLabels.ps1
```diff
@@ -565,6 +628,121 @@ function Set-GitHubLabel
     }
 }
 
+function Add-GitHubLabel
+{
+<#
+    .DESCRIPTION
+        Adds a label to an issue in the given GitHub repository.
```
HowardWolosky: Looks like you're missing the [Replace all labels for an issue](https://developer.github.com/v3/issues/labels/#replace-all-labels-for-an-issue) functionality.

GitHubLabels.ps1
```diff
@@ -565,6 +628,121 @@ function Set-GitHubLabel
...
+
+    .PARAMETER Uri
+        Uri for the repository.
+        The OwnerName and RepositoryName will be extracted from here instead of needing to provide
+        them individually.
+
+    .PARAMETER Issue
+        Issue number to add the label to.
+
+    .PARAMETER LabelName
```
HowardWolosky: We use `Name` as the parameter name for all of the other Label functions, with it _aliased_ to be `LabelName`.

GitHubLabels.ps1
```diff
@@ -565,6 +628,121 @@ function Set-GitHubLabel
...
+        Array of label names to add to the issue
+
+    .PARAMETER Replace
```
HowardWolosky: From an internal implementation perspective, I think this is fine.  I'm not sure I buy into this though from an exported API design perspective.  I think it would generally be frowned-upon from a PowerShell naming perspective to combine both "Add" and "Set"/Replace functionality within the same exported function.

GitHubLabels.ps1
```diff
@@ -565,6 +628,121 @@ function Set-GitHubLabel
...
+    else
+    {
+        $method = 'Post'
+    }
+
+    $params = @{
+        'UriFragment' = "repos/$OwnerName/$RepositoryName/issues/$Issue/labels"
+        'Body' = (ConvertTo-Json -InputObject $hashBody)
+        'Method' = $method
+        'Description' =  "Adding label $Name to issue $Issue in $RepositoryName"
```
HowardWolosky: This definitely can't be working given that you currently don't have a `$Name` parameter.  On top of that, it's not totally accurate if using the `-Replace` switch as currently designed, as it's _replacing_ the labels.

GitHubLabels.ps1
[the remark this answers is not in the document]
HowardWolosky: I _think_ we should stick with `Add-GitHubLabel` and `Remove-GitHubLabel`.  Otherwise, if we followed this newer suggestion, we'd _also_ need `Remove-GitHubMilestonLabel` and that just "feels" weird.

GitHubLabels.ps1
```diff
@@ -287,6 +337,8 @@ function Remove-GitHubLabel
         [Alias('LabelName')]
         [string] $Name,
 
+        [int] $Issue,
```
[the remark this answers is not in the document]
joseartrivera: Couldn't find any docs but the official Octokit library uses an int for issue number throughout the code, as does a few of the other 3rd party repos I skimmed through.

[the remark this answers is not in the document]
HowardWolosky: That is sufficient "specific guidance" for me.  Can you change the current `[sting]` types for the existing functions to be `[int]` as well then?  Thanks for digging into this deeper!

GitHubLabels.ps1
joseartrivera: If I change this to Set-GithubLabel we are going to run into the same ass with Remove-GithubLabel. Setting the title/color/etc of a github label is a lot different than setting the labels on an issue. I think at this point it would be better to have a separate set of Add-GithubIssueLabel/Remove-GithubIssueLabel/Set-GithubIssueLabel commands just for issues. There will not be commands for Milestones since labels put on labels are not directly set on a Milestone but instead computed by all of the labels that are set on issues that are a part of a Milestone.

HowardWolosky: Ok, I'm convinced.  Break 'em out that way.  Thanks!

GitHubLabels.ps1
```diff
@@ -85,9 +108,33 @@ function Get-GitHubLabel
...
+        else
+        {
+            $description = "Getting labels for $RepositoryName"
+        }
+    }
+
     $params = @{
-        'UriFragment' = "repos/$OwnerName/$RepositoryName/labels/$Name"
-        'Description' =  "Getting all labels for $RepositoryName"
+        'UriFragment' = $uriFragment
```
HowardWolosky: I mentioned this in @etgottli's PR as well.  I'm surprised that PSScriptAnalyzer isn't catching this.  `$uriFragment` and `$description` aren't initially defined in the same scope as this.  While PowerShell supports this in theory, it's safer to explicitly define them in the same scope (outside/before your `if/then/else`.

GitHubLabels.ps1
```diff
@@ -652,11 +639,111 @@ function Add-GitHubLabel
...
+        Add-GitHubLabel -OwnerName PowerShell -RepositoryName PowerShellForGitHub -Issue 1 -Name $labels
+
+        Adds labels to an issue in the PowerShellForGitHub project.
+#>
+    [CmdletBinding(
+        SupportsShouldProcess,
+        DefaultParametersetName='Elements')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "", Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
+    param(
+        [Parameter(ParameterSetName='Elements')]
```
danbelcher-MSFT: Are `$OwnerName` and `$RepositoryName` mandatory for their parameter set?

danbelcher-MSFT: Can you double-check the help documentation shows the correct parameter sets for each command you added/modified?

GitHubLabels.ps1
```diff
@@ -711,27 +799,126 @@ function Add-GitHubLabel
...
+        REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
+
+    .PARAMETER NoStatus
+        If this switch is specified, long-running commands will run on the main thread
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .EXAMPLE
+        Remove-GitHubLabel -OwnerName PowerShell -RepositoryName PowerShellForGitHub -Name TestLabel -Issue 1
```
danbelcher-MSFT: Doesn't match the function name

Tests/Config/Settings.ps1
```diff
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 # The account that the tests will be running under
-$script:ownerName = 'PowerShellForGitHubTeam'
+$script:ownerName = 'jorivtest'
```
danbelcher-MSFT: Are these intentionally changed?

Tests/GitHubLabels.tests.ps1
```diff
@@ -296,34 +296,66 @@ if ($script:accessTokenConfigured)
...
+        'enhancement', 'up for grabs', 'question', 'discussion', 'wontfix', 'in progress', 'ready')
+        Add-GitHubIssueLabel -OwnerName $script:ownerName -RepositoryName $repositoryName -Issue $issue.number -LabelName $labelsToAdd
 
+        Context 'For removing individual issues'{
+            Remove-GitHubIssueLabel -OwnerName $script:ownerName -RepositoryName $repositoryName -Name "discussion" -Issue $issue.number
+            Remove-GitHubIssueLabel -OwnerName $script:ownerName -RepositoryName $repositoryName -Name "question" -Issue $issue.number
+            Remove-GitHubIssueLabel -OwnerName $script:ownerName -RepositoryName $repositoryName -Name "bug" -Issue $issue.number
+            $labelIssues = Get-GitHubLabel -OwnerName $script:ownerName -RepositoryName $repositoryName -Issue $issue.number
+    
             It 'Should have added the expected number of labels' {
```
danbelcher-MSFT: The names for these tests are very generic.

Tests/GitHubLabels.tests.ps1
```diff
@@ -272,7 +272,7 @@ if ($script:accessTokenConfigured)
         $null = Remove-GitHubRepository -OwnerName $script:ownerName -RepositoryName $repositoryName
     }
 
-    Describe 'Adding and removing labels to an issue'{
+    Describe 'Adding labels to an issue'{
         $repositoryName = [Guid]::NewGuid().Guid
```
danbelcher-MSFT: I have the usual feedback about leaving `$repositoryName` as just `[Guid]::NewGuid()`

Tests/GitHubLabels.tests.ps1
```diff
@@ -272,7 +272,7 @@ if ($script:accessTokenConfigured)
...
         $null = New-GitHubRepository -RepositoryName $repositoryName
         Set-GitHubLabel -OwnerName $script:ownerName -RepositoryName $repositoryName -Label $script:defaultLabels
```
danbelcher-MSFT: Same feedback about `$script:ownerName -> $ownerName`

Tests/Config/Settings.ps1
joseartrivera: I think you are looking at an older revision.

Tests/GitHubLabels.tests.ps1
joseartrivera: I think you are looking at an older revision.

Tests/GitHubLabels.tests.ps1
joseartrivera: Not sure what you mean by this, any suggestions for better test names?

danbelcher-MSFT: It just seems like if I run the tests from the console and I see a test label like 'Should have added the expected number of labels', I know nothing about the expected behavior. The test labels are as informative as simply saying It 'should do the right thing'.

joseartrivera: Makes sense, I made them a bit more descriptive

GitHubLabels.ps1
```diff
@@ -859,6 +821,114 @@ function Set-GitHubIssueLabel
...
+        [string] $Uri,
+
+        [Parameter(Mandatory)]
+        [int] $Issue,
+
+        [ValidateNotNullOrEmpty()]
+        [Alias('LabelName')]
+        [string] $Name,
+
+        [switch] $RemoveAll,
```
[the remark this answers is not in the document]
joseartrivera: Whoops, it should really just have Name, if Name is not provided then it will remove all.

GitHubLabels.ps1
```diff
@@ -565,6 +616,320 @@ function Set-GitHubLabel
...
+        [string] $Uri,
+
+        [Parameter(Mandatory)]
+        [int] $Issue,
+
+        [Parameter(Mandatory)]
+        [Alias('LabelName')]
+        [string[]] $Name,
+
+        [switch] $Replace,
```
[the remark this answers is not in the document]
joseartrivera: removed

PR #60, opened 2018-12-04
Add `LogRequestBody` configuration option to make development debugging easier

HowardWolosky:
If enabled, the JSON-serialized version of the body of any REST request will
be written to verbose output, so that you can see exactly what is being sent
to GitHub.

### DIFF
--- a/GitHubConfiguration.ps1
+++ b/GitHubConfiguration.ps1
@@ -114,6 +114,10 @@ function Set-GitHubConfiguration
         log entry.  This can be useful if you have concurrent PowerShell sessions all logging
         to the same file, as it would then be possible to filter results based on ProcessId.
 
+    .PARAMETER LogRequestBody
+        If specified, the JSON body of the REST request will be logged to verbose output.
+        This can be helpful for debugging purposes.
+
     .PARAMETER LogTimeAsUtc
         If specified, all times logged will be logged as UTC instead of the local timezone.
 
@@ -179,6 +183,8 @@ function Set-GitHubConfiguration
         [string] $LogPath,
 
         [switch] $LogProcessId,
+
+        [switch] $LogRequestBody,
 
         [switch] $LogTimeAsUtc,
 
@@ -262,6 +268,7 @@ function Get-GitHubConfiguration
             'DisableTelemetry',
             'LogPath',
             'LogProcessId',
+            'LogRequestBody',
             'LogTimeAsUtc',
             'RetryDelaySeconds',
             'SuppressNoTokenWarning',
@@ -593,6 +600,7 @@ function Import-GitHubConfiguration
         'defaultRepositoryName' = [String]::Empty
         'logPath' = $logPath
         'logProcessId' = $false
+        'logRequestBody' = $false
         'logTimeAsUtc' = $false
         'retryDelaySeconds' = 30
         'suppressNoTokenWarning' = $false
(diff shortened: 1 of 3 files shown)


GitHubCore.ps1
```diff
@@ -217,7 +221,7 @@ function Invoke-GHRestMethod
             if ($PSCmdlet.ShouldProcess($jobName, "Start-Job"))
             {
                 [scriptblock]$scriptBlock = {
-                    param($Url, $Method, $Headers, $Body, $ValidBodyContainingRequestMethods, $TimeoutSec, $ScriptRootPath)
+                    param($Url, $Method, $Headers, $Body, $ValidBodyContainingRequestMethods, $TimeoutSec, $LogRequestBody, $ScriptRootPath)
```
danbelcher-MSFT: Out-of-scope for this PR, but instead of maintaining two codepaths, it seems easier to define one scriptblock and either invoke as job or invoke locally.  If there are differences in the two paths, we can separate out what is common and what is not.

GitHubConfiguration.ps1
```diff
@@ -180,6 +184,8 @@ function Set-GitHubConfiguration
 
         [switch] $LogProcessId,
 
+        [switch] $LogRequestBody,
```
danbelcher-MSFT: Instead of adding a new switch, is it possible to just use the existing `-Debug` switch? It would also be better for the module if it made better use of the various output streams so that users could redirect or suppress different levels of information.

GitHubCore.ps1
```diff
@@ -272,7 +280,7 @@ function Invoke-GHRestMethod
                     }
                 }
 
-                $null = Start-Job -Name $jobName -ScriptBlock $scriptBlock -Arg @($url, $Method, $headers, $Body, $ValidBodyContainingRequestMethods, (Get-GitHubConfiguration -Name WebRequestTimeoutSec), $PSScriptRoot)
+                $null = Start-Job -Name $jobName -ScriptBlock $scriptBlock -Arg @($url, $Method, $headers, $Body, $ValidBodyContainingRequestMethods, (Get-GitHubConfiguration -Name WebRequestTimeoutSec), (Get-GitHubConfiguration -Name LogRequestBody), $PSScriptRoot)
```
danbelcher-MSFT: Is it time yet to start splitting these arguments across separate lines?
```powershell
$null = Start-Job -Name $jobName -ScriptBlock $scriptBlock -Arg @(
    $url,
    $Method
    ...
)
```

GitHubCore.ps1
HowardWolosky: Good suggestion.  I've opened up #61 to track that suggestion as a future enhancement,

GitHubConfiguration.ps1
HowardWolosky: This is a great suggestion at a high level, but becomes more complicated when we start dealing with jobs that are executing asynchronously.  I found [this interesting post](https://social.technet.microsoft.com/forums/windowsserver/en-US/696ceda1-ea2a-43ba-b648-9ce058db210b/how-to-debug-a-job-created-using-startjob) about debugging an executing Job, but it looks like this would end up complicating the code quite a bit further.  I think for now, inspection via an optional log/verbose output helps satisfy this direct need, and we can look at adding `Write-Log -Level Debug` messages in other places on an as-is needed basis in the future.

GitHubCore.ps1
HowardWolosky: At this point, I'd agree. Will fix.

PR #64, opened 2018-12-06
Add support for Issue Events Github API

joseartrivera:
Add wrappers for the [Issue Events](https://developer.github.com/v3/issues/events/) APIs


GitHubEvents.ps1
```diff
@@ -0,0 +1,119 @@
...
+    {
+        $uriFragment = "repos/$OwnerName/$RepositoryName/issues/events"
+        $description = "Getting events for $RepositoryName"
+    }
+
+    $params = @{
+        'UriFragment' = $uriFragment
+        'Description' = $description
+        'AccessToken' = $AccessToken
+        'AcceptHeader' = 'application/vnd.github.starfox-preview+json,application/vnd.github.sailor-v-preview+json,application/vnd.github.symmetra-preview+json,application/vnd.github.machine-man-preview'
```
danbelcher-MSFT: Long line.  You can specify an array of values and then `-join` them here:
```powershell
$acceptHeaders = @(
    'application/vnd.github.starfox-preview+json',
    'application/vnd.github.sailer-v-preview+json',
    ...
)

'AcceptHeader' = $acceptHeaders -join ','
```

Tests/GitHubEvents.tests.ps1
```diff
@@ -0,0 +1,159 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+<#
+.Synopsis
+   Tests for GitHubLabels.ps1 module
```
danbelcher-MSFT: Events?

Tests/GitHubEvents.tests.ps1
```diff
@@ -0,0 +1,159 @@
...
+# Backup the user's configuration before we begin, and ensure we're at a pure state before running
+# the tests.  We'll restore it at the end.
+$configFile = New-TemporaryFile
+
+try
+{
+    Backup-GitHubConfiguration -Path $configFile
+    Reset-GitHubConfiguration
+
+    if ($script:accessTokenConfigured)
```
danbelcher-MSFT: Why not check for this before having to touch the users' config file?  Why not exit immediately if there is no configured access token?  And why not throw an exception if that is the case?

Tests/GitHubEvents.tests.ps1
```diff
@@ -0,0 +1,159 @@
...
+    {
+        Describe 'Getting events from repository' {
+            $repositoryName = [Guid]::NewGuid().Guid
+            $null = New-GitHubRepository -RepositoryName $repositoryName
+
+            Context 'For getting events from a new repository' {
+                $events = @(Get-GitHubEvent -OwnerName $script:ownerName -RepositoryName $repositoryName)
```
danbelcher-MSFT: This feedback applies throughout the file, but I don't think it's that meaningful to use the explicit script scope here.  The reason is that this variable is already being accessed from script scope, so this variable is just as much script scope as any other variable here, e.g. `$repositoryName`.  Having explicit script scope syntax makes sense when you are inside a function's local scope and referring to a variable that is outside of your local scope, i.e. reaching above your current scope to a variable defined out in the script.  Here, in this scope, everything is script-scope.

Tests/GitHubEvents.tests.ps1
```diff
@@ -0,0 +1,159 @@
...
+
+                It 'Should have no events' {
+                    $($events).Count | Should be 0
```
danbelcher-MSFT: Sub-expression operator `$()` doesn't add anything here.  `$events` is already an array and `$($events)` just evaluates as `$events` anyways.  `$events.Count` is the same.

Tests/GitHubEvents.tests.ps1
```diff
@@ -0,0 +1,159 @@
...
+
+try
+{
+    Backup-GitHubConfiguration -Path $configFile
+    Reset-GitHubConfiguration
+
+    if ($script:accessTokenConfigured)
+    {
+        Describe 'Getting events from repository' {
+            $repositoryName = [Guid]::NewGuid().Guid
```
danbelcher-MSFT: Generally, you want to keep objects as objects, and not as strings in PowerShell.  That would mean `$repositoryName = [Guid]::NewGuid()`.  Then `$repositoryName` is a `[Guid]` object.  A `[Guid]` is implicitly convertible to `[string]`, therefore, these statements are the same:
```powershell
[string] [Guid]::NewGuid()
...
[Guid]::NewGuid().Guid
```

What this means for your code is that you can keep `$repostitoryName` as a `[Guid]` and when you pass that `[Guid]` to `-RepositoryName` parameter which is a `[string]`, the `[Guid]` will be implicitly converted as you would expect.

Tests/GitHubEvents.tests.ps1
```diff
@@ -0,0 +1,159 @@
...
+            Context 'For getting events from a new issue' {
+                $events = @(Get-GitHubEvent -OwnerName $script:ownerName -RepositoryName $repositoryName -Issue $issue.number)
+
+                It 'Should have no events' {
+                    $($events).Count | Should be 0
+                }
+            }
+
+            Update-GitHubIssue -OwnerName $script:ownerName -RepositoryName $repositoryName -Issue $issue.number -State closed
+            Update-GitHubIssue -OwnerName $script:ownerName -RepositoryName $repositoryName -Issue $issue.number -State open
```
danbelcher-MSFT: Doesn't this belong as part of the below context?

Tests/GitHubEvents.tests.ps1
joseartrivera: I am guessing that the tests were set up this way since some tests wouldn't necessarily need an accessToken to complete the tests while the ones that did would fail during PR builds? @HowardWolosky any ideas?

GitHubEvents.ps1
```diff
@@ -0,0 +1,125 @@
...
+        [string] $RepositoryName,
+
+        [Parameter(Mandatory, ParameterSetName='RepositoryUri')]
+        [Parameter(Mandatory, ParameterSetName='IssueUri')]
+        [Parameter(Mandatory, ParameterSetName='EventUri')]
+        [string] $Uri,
+
+        [Parameter(Mandatory, ParameterSetName='EventUri')]
+        [Parameter(Mandatory, ParameterSetName='EventElements')]
+        [string] $EventID,
```
HowardWolosky: I think this should be `Event` to be consistent with how we name the other parameters.

HowardWolosky: It looks like this should be an `[int]` and not a `[string]`

GitHubEvents.ps1
```diff
@@ -0,0 +1,125 @@
...
+        $description = "Getting events for issue $Issue in $RepositoryName"
+    }
+    else
+    {
+        $uriFragment = "repos/$OwnerName/$RepositoryName/issues/events"
+        $description = "Getting events for $RepositoryName"
+    }
+
+    $acceptHeaders = @(
+    'application/vnd.github.starfox-preview+json',
```
HowardWolosky: Can you do single-level indentation for these?

GitHubEvents.ps1
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    if ($PSBoundParameters.ContainsKey('EventID'))
+    {
+        $uriFragment = "repos/$OwnerName/$RepositoryName/issues/events/$EventID"
+        $description = "Getting event $EventID for $RepositoryName"
+    }
+    elseif ($PSBoundParameters.ContainsKey('Issue'))
+    {
+        $uriFragment = "repos/$OwnerName/$RepositoryName/issues/$Issue/events`?"
```
HowardWolosky: I don't think there are any get parameters, so no need for the
```powershell
`?
```

GitHubEvents.ps1
danbelcher-MSFT: We chose not to use `$Event` because it is an existing PowerShell automatic variable.

Tests/GitHubEvents.tests.ps1
HowardWolosky: That's a carry-over from the test files from before I took over the repo.  I left it in, because many r/o scenarios don't require an access token at all.  Upon re-examination, the message is slightly off (since tests aren't skipped, they will just fail as a side-effect).  We could, in theory, just immediately throw if no access token isn't provided, but again, an access token isn't strictly necessary.

PR #83, opened 2019-01-11
Add ability to specify a MediaType for issues

joseartrivera:
This PR adds the ability to specify the MediaType in which the Issues APIs will return the body of the issue (raw, text, html, full). It is consistent with the capabilities currently available in the Comments APIs.


GitHubComments.ps1
```diff
@@ -181,7 +181,7 @@ function Get-GitHubComment
         'UriFragment' = $uriFragment
         'Description' = $description
         'AccessToken' = $AccessToken
-        'AcceptHeader' = (Get-MediaAcceptHeader -MediaType $MediaType)
+        'AcceptHeader' = (Get-MediaAcceptHeader -MediaType $MediaType -AcceptHeader 'application/vnd.github.squirrel-girl-preview')
```
HowardWolosky: Seems like it would make sense to have a `$script`-level constant for `application/vnd.github.squirrel-girl-preview`, and then we can start making constants for others in a later change.

Issue #84, opened 2019-01-12
Repositories - Merging


iamraghavs:
Issue to track the implementation of the APIs for "merging" under the "Repository" section in #70 
API reference: https://developer.github.com/v3/repos/merging/

I will update this issue with my proposals for the cmdlet names and parameters

HowardWolosky:
Thanks @raghav710!  Overall, this looks great.  Looking forward to having this functionality.

Some minor feedback here:

* Please keep `AccessToken` and `NoStatus` as the _last_ parameters in the method, just for consistency with the rest of the module.
* You may want to call `Invoke-GHRestMethod` with the `-ExtendedResult` parameter so that you can detect a `201` vs a `204` and be able to do `Write-Log -Level Warning` in the event that it's a `204` explaining that there's nothing to merge.

raghav710:
@HowardWolosky thanks for the note on the `201` vs `204`. Looking forward to getting this in :)

HowardWolosky:
Me too.

The test may be hard to write though because we don't yet have API support for creating new pull requests... Something to keep in mind.

raghav710:
@HowardWolosky I think I need the ability to create branches so I can add tests for this. Shall I concentrate on adding that and then come back to this?

raghav710:
Also, here is a tentative PR: https://github.com/PowerShell/PowerShellForGitHub/pull/89
I've tested the code manually using my own access token

HowardWolosky:
Thanks Raghav!  I've given your tentative PR an initial code review. I do think that it would be best to only check in features that have corresponding tests that can help ensure that the module remains working as expected.  If you're able to add the necessary, tested methods for creating branches so that you can add tests for this PR, that would be great.

I think you may run into some trouble writing tests for the [refs API's](https://developer.github.com/v3/git/refs/) too though, because per the documentation, you can't create new refs on empty repos...so, you'd have to make sure that it has content in it first.  It's possible that using `-AutoInit` to create it with an empty README will be sufficient to satisfy that requirement however.

PR #97, opened 2019-03-24
Creating and retrieving references

raghav710:
* Tests for a few common scenarios
* Only Get and Create references implemented as of now as it is required for #84 
* PSScriptAnalyser Run returns no issues

### DIFF
--- a/GitHubReferences.ps1
+++ b/GitHubReferences.ps1
@@ -1,0 +1,219 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+function Get-GitHubReference
+{
+<#
+    .SYNOPSIS
+        Retrieve a reference of a given GitHub repository.
+
+    .DESCRIPTION
+        Retrieve a reference of a given GitHub repository.
+        The Git repo for this module can be found here: http://aka.ms/PowerShellForGitHub
+
+    .PARAMETER OwnerName
+        Owner of the repository.
+        If not supplied here, the DefaultOwnerName configuration property value will be used.
+
+    .PARAMETER RepositoryName
+        Name of the repository.
+        If not supplied here, the DefaultRepositoryName configuration property value will be used.
+
+    .PARAMETER Uri
+        Uri for the repository.
+        The OwnerName and RepositoryName will be extracted from here instead of needing to provide
+        them individually.
+
+    .PARAMETER Reference
+        Name of the reference, for example: "heads/<branch name>" for branches and "tags/<tag name>" for tags
+
+    .PARAMETER AccessToken
+        If provided, this will be used as the AccessToken for authentication with the
+        REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
+
+    .PARAMETER NoStatus
+        If this switch is specified, long-running commands will run on the main thread
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .EXAMPLE
+        Get-GitHubReference -OwnerName Powershell -RepositoryName PowerShellForGitHub -Reference heads/master
+#>
+    [CmdletBinding(
+        SupportsShouldProcess,
+        DefaultParametersetName='Elements')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "", Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
+    param(
+        [Parameter(ParameterSetName='Elements')]
+        [string] $OwnerName,
+
+        [Parameter(ParameterSetName='Elements')]
+        [string] $RepositoryName,
+
+        [Parameter(
+            Mandatory,
+            ParameterSetName='Uri')]
+        [string] $Uri,
+
+        [Parameter(Mandatory)]
+        [string] $Reference,
+
+        [string] $AccessToken,
+
+        [switch] $NoStatus
+    )
+
+    Write-InvocationLog -Invocation $MyInvocation
+
+    $elements = Resolve-RepositoryElements -BoundParameters $PSBoundParameters -DisableValidation
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    if ($OwnerName -xor $RepositoryName)
+    {
+        $message = 'You must specify both Owner Name and Repository Name.'
+        Write-Log -Message $message -Level Error
+        throw $message
+    }
+    $uriFragment = "/repos/$OwnerName/$RepositoryName/git/refs/$Reference"
+    $description =  "Getting Reference $Reference for $RepositoryName"
+
+    $params = @{
+        'UriFragment' = $uriFragment
+        'Method' = 'Get'
+        'Description' =  $description
+        'AcceptHeader' = 'application/vnd.github.symmetra-preview+json'
+        'AccessToken' = $AccessToken
+        'TelemetryEventName' = $MyInvocation.MyCommand.Name
+        'TelemetryProperties' = $telemetryProperties
+        'NoStatus' = (Resolve-ParameterWithDefaultConfigurationValue -Name NoStatus -ConfigValueName DefaultNoStatus)
+    }
+
+    try
+    {
+        return Invoke-GHRestMethod @params
+    }
+    catch
+    {
+        Write-InteractiveHost "No reference named $Reference exists in repository $RepositoryName" -NoNewline -f Red
+        Write-Log -Level Error "No reference named $Reference exists in repository $RepositoryName"
+        return $null
+    }
+
+}
+
+function New-GitHubReference
+{
+    <#
+    .SYNOPSIS
+        Create a reference in a given GitHub repository.
+
+    .DESCRIPTION
+        Create a reference in a given GitHub repository.
+        The Git repo for this module can be found here: http://aka.ms/PowerShellForGitHub
+
+    .PARAMETER OwnerName
+        Owner of the repository.
+        If not supplied here, the DefaultOwnerName configuration property value will be used.
+
+    .PARAMETER RepositoryName
+        Name of the repository.
+        If not supplied here, the DefaultRepositoryName configuration property value will be used.
+
+    .PARAMETER Uri
+        Uri for the repository.
+        The OwnerName and RepositoryName will be extracted from here instead of needing to provide
+        them individually.
+
+    .PARAMETER Reference
+        The name of the fully qualified reference to be created (eg: refs/heads/master)
+
+    .PARAMETER Sha
+        The SHA1 value for the reference to be created
+
+    .PARAMETER AccessToken
+        If provided, this will be used as the AccessToken for authentication with the
+        REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
+
+    .PARAMETER NoStatus
+        If this switch is specified, long-running commands will run on the main thread
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .EXAMPLE
+        -GitHubReference -OwnerName Powershell -RepositoryName PowerShellForGitHub -Reference refs/heads/master -Sha aa218f56b14c9653891f9e74264a383fa43fefbd
+    #>
+    [CmdletBinding(
+        SupportsShouldProcess,
+        DefaultParametersetName='Elements')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "", Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
+    param(
+        [Parameter(ParameterSetName='Elements')]
+        [string] $OwnerName,
+
+        [Parameter(ParameterSetName='Elements')]
+        [string] $RepositoryName,
+
+        [Parameter(
+            Mandatory,
+            ParameterSetName='Uri')]
+        [string] $Uri,
+
+        [Parameter(Mandatory)]
+        [string] $Reference,
+
+        [Parameter(Mandatory)]
+        [string] $Sha,
+
+        [string] $AccessToken,
+
+        [switch] $NoStatus
+    )
+
+    Write-InvocationLog -Invocation $MyInvocation
+
+    $elements = Resolve-RepositoryElements -BoundParameters $PSBoundParameters -DisableValidation
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    if ($OwnerName -xor $RepositoryName)
+    {
+        $message = 'You must specify both Owner Name and Repository Name.'
+        Write-Log -Message $message -Level Error
+        throw $message
+    }
+
+    $uriFragment = "/repos/$OwnerName/$RepositoryName/git/refs"
+    $description =  "Creating Reference $Reference for $RepositoryName from SHA $Sha"
+
+    $hashBody = @{
+        'ref' = $Reference
+        'sha' = $Sha
+    }
+
+    $params = @{
+        'UriFragment' = $uriFragment
+        'Method' = 'Post'
+        'Body' = (ConvertTo-Json -InputObject $hashBody)
+        'Description' =  $description
+        'AcceptHeader' = 'application/vnd.github.symmetra-preview+json'
+        'AccessToken' = $AccessToken
+        'TelemetryEventName' = $MyInvocation.MyCommand.Name
+        'TelemetryProperties' = $telemetryProperties
+        'NoStatus' = (Resolve-ParameterWithDefaultConfigurationValue -Name NoStatus -ConfigValueName DefaultNoStatus)
+    }
+
+    return Invoke-GHRestMethod @params
+}
--- a/PowerShellForGitHub.psd1
+++ b/PowerShellForGitHub.psd1
@@ -32,6 +32,7 @@
         'GitHubMiscellaneous.ps1',
         'GitHubOrganizations.ps1',
         'GitHubPullRequests.ps1',
+        'GitHubReferences.ps1',
         'GitHubRepositories.ps1',
         'GitHubRepositoryForks.ps1',
         'GitHubRepositoryTraffic.ps1',
@@ -115,7 +116,9 @@
         'Update-GitHubCurrentUser',
         'Update-GitHubIssue',
         'Update-GitHubLabel',
-        'Update-GitHubRepository'
+        'Update-GitHubRepository',
+        'Get-GitHubReference',
+        'New-GithubReference'
     )
 
     AliasesToExport = @(
--- a/Tests/GitHubReferences.Tests.ps1
+++ b/Tests/GitHubReferences.Tests.ps1
@@ -1,0 +1,173 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+<#
+.Synopsis
+   Tests for GitHubReferences.ps1 module
+#>
+
+[String] $root = Split-Path -Parent (Split-Path -Parent $MyInvocation.MyCommand.Path)
+. (Join-Path -Path $root -ChildPath 'Tests\Config\Settings.ps1')
+Import-Module -Name $root -Force
+
+function Initialize-AppVeyor
+{
+<#
+    .SYNOPSIS
+        Configures the tests to run with the authentication information stored in AppVeyor
+        (if that information exists in the environment).
+
+    .DESCRIPTION
+        Configures the tests to run with the authentication information stored in AppVeyor
+        (if that information exists in the environment).
+
+        The Git repo for this module can be found here: http://aka.ms/PowerShellForGitHub
+
+    .NOTES
+        Internal-only helper method.
+
+        The only reason this exists is so that we can leverage CodeAnalysis.SuppressMessageAttribute,
+        which can only be applied to functions.
+
+        We call this immediately after the declaration so that AppVeyor initialization can heppen
+        (if applicable).
+
+#>
+    [CmdletBinding()]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingConvertToSecureStringWithPlainText", "", Justification="Needed to configure with the stored, encrypted string value in AppVeyor.")]
+    param()
+
+    if ($env:AppVeyor)
+    {
+        $secureString = $env:avAccessToken | ConvertTo-SecureString -AsPlainText -Force
+        $cred = New-Object System.Management.Automation.PSCredential "<username is ignored>", $secureString
+        Set-GitHubAuthentication -Credential $cred
+
+        $script:ownerName = $env:avOwnerName
+        $script:organizationName = $env:avOrganizationName
+
+        $message = @(
+            'This run is executed in the AppVeyor environment.',
+            'The GitHub Api Token won''t be decrypted in PR runs causing some tests to fail.',
+            '403 errors possible due to GitHub hourly limit for unauthenticated queries.',
+            'Use Set-GitHubAuthentication manually. modify the values in Tests\Config\Settings.ps1,',
+            'and run tests on your machine first.')
+        Write-Warning -Message ($message -join [Environment]::NewLine)
+    }
+}
+
+Initialize-AppVeyor
+
+$accessTokenConfigured = Test-GitHubAuthenticationConfigured
+if (-not $accessTokenConfigured)
+{
+    $message = @(
+        'GitHub API Token not defined, some of the tests will be skipped.',
+        '403 errors possible due to GitHub hourly limit for unauthenticated queries.')
+    Write-Warning -Message ($message -join [Environment]::NewLine)
+}
+
+# Backup the user's configuration before we begin, and ensure we're at a pure state before running
+# the tests.  We'll restore it at the end.
+$configFile = New-TemporaryFile
+
+try
+{
+    Backup-GitHubConfiguration -Path $configFile
+    Reset-GitHubConfiguration
+    Set-GitHubConfiguration -DisableTelemetry # We don't want UT's to impact telemetry
+    Set-GitHubConfiguration -LogRequestBody # Make it easier to debug UT failures
+
+    if ($accessTokenConfigured)
+    {
+
+        Describe 'Create a new reference(branch) in repository' {
+            $repositoryName = [Guid]::NewGuid()
+            $repo = New-GitHubRepository -RepositoryName $repositoryName -AutoInit
+            $existingref = @(Get-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName -Reference "heads/master")
+            $sha = $existingref.object.sha
+
+            Context 'On creating a valid reference in a new repository from a given SHA' {
+                $refName = "refs/heads/" + [Guid]::NewGuid().ToString()
+                $result = @(New-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName -Reference $refName -Sha $sha)
+
+                It 'Should successfully create the reference' {
+                    $result.ref | Should Be $refName
+                }
+            }
+
+            Context 'On creating a valid reference in a new repository (specified by Uri) from a given SHA' {
+                $refName = "refs/heads/" + [Guid]::NewGuid().ToString()
+                $result = @(New-GitHubReference -Uri $repo.svn_url -Reference $refName -Sha $sha)
+
+                It 'Should successfully create the reference' {
+                    $result.ref | Should Be $refName
+                }
+            }
+
+            Context 'On creating an existing reference in a new repository from a given SHA' {
+                $refName = "refs/heads/master"
+
+                It 'Should throw an Exception' {
+                    { @(New-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName -Reference $refName -Sha $sha) } | Should Throw
+                }
+            }
+
+            Context 'On creating an existing reference in a new repository (specified by Uri) from a given SHA' {
+                $refName = "refs/heads/master"
+
+                It 'Should throw an exception' {
+                    { @(New-GitHubReference -Uri $repo.svn_url -Reference $refName -Sha $sha) }  | Should Throw
+                }
+            }
+
+
+
+            $null = Remove-GitHubRepository -OwnerName $ownerName -RepositoryName $repositoryName
+        }
+
+        Describe 'Getting a reference(branch) from repository' {
+            $repositoryName = [Guid]::NewGuid()
+            $repo = New-GitHubRepository -RepositoryName $repositoryName -AutoInit
+
+            Context 'On getting a valid reference from a new repository' {
+                $reference = @(Get-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName -Reference "heads/master")
+
+                It 'Should return details of the reference' {
+                    $reference.ref | Should be "refs/heads/master"
+                }
+            }
+
+            Context 'On getting an invalid reference from a new repository' {
+                $reference = @(Get-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName -Reference "heads/someRandomRef")
+
+                It 'Should not return any details' {
+                    $reference | Should be $null
+                }
+            }
+
+            Context 'On getting a valid reference using Uri from a new repository' {
+                $reference = @(Get-GitHubReference -Uri $repo.svn_url -Reference "heads/master")
+
+                It 'Should return details of the reference' {
+                    $reference.ref | Should be "refs/heads/master"
+                }
+            }
+
+            Context 'On getting an invalid reference using Uri from a new repository' {
+                $reference = @(Get-GitHubReference -Uri $repo.svn_url -Reference "heads/someRandomRef")
+
+                It 'Should not return any details' {
+                    $reference | Should be $null
+                }
+            }
+            $null = Remove-GitHubRepository -OwnerName $ownerName -RepositoryName $repositoryName
+        }
+    }
+}
+catch
+{
+    # Restore the user's configuration to its pre-test state
+    Restore-GitHubConfiguration -Path $configFile
+}
+


GitHubReferences.ps1
```diff
@@ -0,0 +1,219 @@
...
+        'Body' = (ConvertTo-Json -InputObject $hashBody)
+        'Description' =  $description
+        'AcceptHeader' = 'application/vnd.github.symmetra-preview+json'
+        'AccessToken' = $AccessToken
+        'TelemetryEventName' = $MyInvocation.MyCommand.Name
+        'TelemetryProperties' = $telemetryProperties
+        'NoStatus' = (Resolve-ParameterWithDefaultConfigurationValue -Name NoStatus -ConfigValueName DefaultNoStatus)
+    }
+
+    return Invoke-GHRestMethod @params
```
raghav710: @HowardWolosky this throws a 422 exception if the reference already exists. I was not sure based on other files on what the expected behaviour should be and am hence allowing it to throw the exception (The exception message contains info about why it failed) and am asserting the same in the tests

Tests/GitHubReferences.Tests.ps1
```diff
@@ -0,0 +1,173 @@
...
+
+        We call this immediately after the declaration so that AppVeyor initialization can heppen
+        (if applicable).
+
+#>
+    [CmdletBinding()]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingConvertToSecureStringWithPlainText", "", Justification="Needed to configure with the stored, encrypted string value in AppVeyor.")]
+    param()
+
+    if ($env:AppVeyor)
```
raghav710: I took this portion verbatim from other tests. I would be happy to move this Appveyor bit to a common util file which is then called from other tests. Thoughts?

raghav710:
I'm planning to rebase my #84 branch on top of this so I can get that ready for review as well while this is being reviewed

PowerShellForGitHub.psd1
```diff
@@ -115,7 +116,9 @@
         'Update-GitHubCurrentUser',
         'Update-GitHubIssue',
         'Update-GitHubLabel',
-        'Update-GitHubRepository'
+        'Update-GitHubRepository',
+        'Get-GitHubReference',
```
HowardWolosky: Please keep these alphabetical.

Tests/GitHubReferences.Tests.ps1
HowardWolosky: Others have mentioned that idea as well in previous PR's, but no-one ever got around to it.  I think it's a solid idea that I completely agree with, but it should be a different PR and not part of this one.  By all means, you're welcome to make that change as well, but please keep that as a separate PR.

GitHubReferences.ps1
```diff
@@ -0,0 +1,227 @@
...
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    if ($OwnerName -xor $RepositoryName)
+    {
+        $message = 'You must specify both Owner Name and Repository Name.'
+        Write-Log -Message $message -Level Error
+        throw $message
+    }
+    $uriFragment = "/repos/$OwnerName/$RepositoryName/git/refs/$Reference"
```
HowardWolosky: nit: line-feed after a closing statement (`}`)

GitHubReferences.ps1
```diff
@@ -0,0 +1,227 @@
...
+        [Parameter(ParameterSetName='Elements')]
+        [string] $RepositoryName,
+
+        [Parameter(
+            Mandatory,
+            ParameterSetName='Uri')]
+        [string] $Uri,
+
+        [Parameter(Mandatory)]
+        [string] $Reference,
```
HowardWolosky: If you make this optional (leaving off `[Parameter(Mandatory)]`, then this will handle _both_ a [single reference](https://developer.github.com/v3/git/refs/#get-a-reference) and [all references](https://developer.github.com/v3/git/refs/#get-all-references), but you'd also have to call `Invoke-GHRestMethodMultipleResult` and branch on what you set `$description` to.  This is a common pattern you can see in other methods (like `Get-GitHubIssue`)

GitHubReferences.ps1
```diff
@@ -0,0 +1,227 @@
...
+
+    Write-InvocationLog -Invocation $MyInvocation
+
+    $elements = Resolve-RepositoryElements -BoundParameters $PSBoundParameters -DisableValidation
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
```
HowardWolosky: Assuming you make the change to have this support _both_ individual and _multiple_ lookup, it would be helpful here to have a boolean property to track if the user provided a `Reference`, like this:

https://github.com/PowerShell/PowerShellForGitHub/blob/68fa5a563b96490774e9c5fae38193194f765157/GitHubIssues.ps1#L190

GitHubReferences.ps1
```diff
@@ -0,0 +1,227 @@
...
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .OUTPUTS
+    [PSCustomObject]
+    Details of the git reference created. Throws an Exception if the reference already exists
+
+    .EXAMPLE
+        -GitHubReference -OwnerName Powershell -RepositoryName PowerShellForGitHub -Reference refs/heads/master -Sha aa218f56b14c9653891f9e74264a383fa43fefbd
```
HowardWolosky: `-GitHubReference` -- typo (missing `New`)

GitHubReferences.ps1
```diff
@@ -0,0 +1,227 @@
...
+        REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
+
+    .PARAMETER NoStatus
+        If this switch is specified, long-running commands will run on the main thread
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .OUTPUTS
+    [PSCustomObject]
```
HowardWolosky: nit: indent

GitHubReferences.ps1
```diff
@@ -0,0 +1,227 @@
...
+        $message = 'You must specify both Owner Name and Repository Name.'
+        Write-Log -Message $message -Level Error
+        throw $message
+    }
+
+    $uriFragment = "/repos/$OwnerName/$RepositoryName/git/refs"
+    $description =  "Creating Reference $Reference for $RepositoryName from SHA $Sha"
+
+    $hashBody = @{
+        'ref' = $Reference
```
HowardWolosky: this should prepend `refs/`, users shouldn't have to deal with that part of the reference.

GitHubReferences.ps1
```diff
@@ -0,0 +1,227 @@
...
+        'sha' = $Sha
+    }
+
+    $params = @{
+        'UriFragment' = $uriFragment
+        'Method' = 'Post'
+        'Body' = (ConvertTo-Json -InputObject $hashBody)
+        'Description' =  $description
+        'AcceptHeader' = 'application/vnd.github.symmetra-preview+json'
```
HowardWolosky: This preview header isn't valid for this API.

GitHubReferences.ps1
```diff
@@ -0,0 +1,227 @@
...
+        throw $message
+    }
+    $uriFragment = "/repos/$OwnerName/$RepositoryName/git/refs/$Reference"
+    $description =  "Getting Reference $Reference for $RepositoryName"
+
+    $params = @{
+        'UriFragment' = $uriFragment
+        'Method' = 'Get'
+        'Description' =  $description
+        'AcceptHeader' = 'application/vnd.github.symmetra-preview+json'
```
HowardWolosky: This preview header isn't valid for this API.

GitHubReferences.ps1
HowardWolosky: Exception when already exists seems fine to me.  Ok to leave as-is.

Tests/GitHubReferences.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+    Set-GitHubConfiguration -DisableTelemetry # We don't want UT's to impact telemetry
+    Set-GitHubConfiguration -LogRequestBody # Make it easier to debug UT failures
+
+    if ($accessTokenConfigured)
+    {
+
+        Describe 'Create a new reference(branch) in repository' {
+            $repositoryName = [Guid]::NewGuid()
+            $repo = New-GitHubRepository -RepositoryName $repositoryName -AutoInit
+            $existingref = @(Get-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName -Reference "heads/master")
```
HowardWolosky: Why are you wrapping the result in array syntax? `@()`
Same question for here and elsewhere in this file.

Tests/GitHubReferences.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+
+            $null = Remove-GitHubRepository -OwnerName $ownerName -RepositoryName $repositoryName
+        }
+
+        Describe 'Getting a reference(branch) from repository' {
+            $repositoryName = [Guid]::NewGuid()
+            $repo = New-GitHubRepository -RepositoryName $repositoryName -AutoInit
+
+            Context 'On getting a valid reference from a new repository' {
+                $reference = @(Get-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName -Reference "heads/master")
```
HowardWolosky: it would probably be worthwhile to have a constant for `head/master` that you use here and elsewhere, and then when testing against it, you could just prepend `refs/` to that same value.

Tests/GitHubReferences.Tests.ps1
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            Context 'On getting an invalid reference using Uri from a new repository' {
+                $reference = @(Get-GitHubReference -Uri $repo.svn_url -Reference "heads/someRandomRef")
+
+                It 'Should not return any details' {
+                    $reference | Should be $null
+                }
+            }
+            $null = Remove-GitHubRepository -OwnerName $ownerName -RepositoryName $repositoryName
```
HowardWolosky: nit: line-feed after a closing statement (`}`)

GitHubReferences.ps1
raghav710: Thanks for the note, I'll have a look at the other methods and adapt the same approach here

raghav710:
@HowardWolosky dropping a note to update that I'll take a look at this probably this weekend. (I wish I could spend more time on this repo)
Thanks for being patient!

HowardWolosky:
Thanks for the update. No worries.  I'll keep this PR open as long as is needed.  I appreciate the contribution.

raghav710:
@HowardWolosky  no need to review this yet, will make a few more changes and address remaining comments before its ready for review again. Thanks for being patient, happy to be back!

Tests/GitHubReferences.Tests.ps1
raghav710: I probably saw this elsewhere and adopted it here. removed it now

GitHubReferences.ps1
```diff
@@ -0,0 +1,241 @@
...
+        Details of the git reference in the given repository
+
+    .EXAMPLE
+        Get-GitHubReference -OwnerName Powershell -RepositoryName PowerShellForGitHub -Reference heads/master
+#>
+    [CmdletBinding(
+        SupportsShouldProcess,
+        DefaultParametersetName='Elements')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "",
+    Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
```
raghav710: Moved this to the next line because this was too far away from the 100 column limit. I haven't seen this being done in other files though. Should I keep it the old way to make it consistent? Thoughts?

Tests/GitHubReferences.Tests.ps1
```diff
@@ -0,0 +1,200 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
raghav710: This file has been refactored a bit, mostly with respect to usage of names like "heads/master" which have been replaced with variables

Tests/GitHubReferences.Tests.ps1
```diff
@@ -0,0 +1,201 @@
...
+            $existingref = Get-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName -Reference $masterRefName
+            $sha = $existingref.object.sha
+            New-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName -Reference $secondRefName -Sha $sha
+            $refNames = @("refs/$masterRefName", "refs/$secondRefName")
+
+            Context 'On getting all references from a new repository' {
+                $reference = Get-GitHubReference -OwnerName $ownerName -RepositoryName $repositoryName
+
+                It 'Should return all references' {
+                    ($reference.ref | Where-Object {$refNames -Contains $_}).Count | Should be $refNames.Count
```
raghav710: This felt like a little too much programming for a test case, but I wasn't happy with checking for the count alone either. Should I just manually check if both items are present or is this fine?

raghav710:
@HowardWolosky this is now ready for another round of review. I havent added the Update and delete bits yet and plan to do so once I'm done with my other Pull request, so this is kind of minimal

Tests/GitHubReferences.Tests.ps1
HowardWolosky: I did actually do this in a later change, so there will be some merging/resolving that you'll need to do as part of this PR.

Tests/GitHubReferences.Tests.ps1
```diff
@@ -0,0 +1,201 @@
...
+            }
+
+            $null = Remove-GitHubRepository -OwnerName $ownerName -RepositoryName $repositoryName
+        }
+
+        Describe 'Getting a reference(branch) from repository' {
+            $repositoryName = [Guid]::NewGuid()
+            $repo = New-GitHubRepository -RepositoryName $repositoryName -AutoInit
+            $masterRefName = "heads/master"
+            $randomRefName = "heads/someRandomRef"
```
HowardWolosky:
```suggestion
            $randomRefName = "heads/$([Guid]::NewGuid())"
```

This way the tests are truly robust.

Tests/GitHubReferences.Tests.ps1
HowardWolosky: I think this is fine.  Thanks for checking.

GitHubReferences.ps1
```diff
@@ -0,0 +1,242 @@
...
+    }
+    catch
+    {
+        if ($PSBoundParameters.ContainsKey('Reference'))
+        {
+            Write-InteractiveHost "No reference named $Reference exists in repository $RepositoryName" -NoNewline -f Red
+            Write-Log -Level Error "No reference named $Reference exists in repository $RepositoryName"
+            return $null
+        }
+        Write-InteractiveHost "No references exist in repository $RepositoryName" -NoNewline -f Red
```
HowardWolosky: In general, I think I want to avoid having individual functions using `Write-InteractiveHost`.  Writing it to `Error` via `Write-Log` should be enough to communicate to users, then `throw`.

GitHubReferences.ps1
```diff
@@ -0,0 +1,242 @@
...
+        'NoStatus' = (Resolve-ParameterWithDefaultConfigurationValue -Name NoStatus -ConfigValueName DefaultNoStatus)
+    }
+
+    try
+    {
+        return Invoke-GHRestMethodMultipleResult @params
+    }
+    catch
+    {
+        if ($PSBoundParameters.ContainsKey('Reference'))
```
HowardWolosky: I don't think you should be doing this additional error checking at all.  If you just let the natural error flow occur (e.g. just `return Invoke-GHRestMethodMultipleResult @params`) and don't bother putting it within this `try/catch`.

raghav710:
@HowardWolosky thanks for the comments. I'll make the required changes and submit for review

GitHubReferences.ps1
raghav710: @HowardWolosky looks like the API/docs have been updated and the single/all reference links are no longer available. Can I understand getting a reference to mean _getting a single reference_ and getting all references to mean _listing matching references_? In that case should I create a new method for listing matching references?

raghav710:
@HowardWolosky I've incorporated your comments and added the update and delete methods. This is ready for another round of review
However I wasn't able to test the update method's positive scenario and would appreciate any help on the same.

I tried creating a couple of branches in a sample repo (https://github.com/raghav710/61603191-d09d-4a1a-8a52-88e719371d8c/pulls) and tried to update the SHA of one of the branches to that of the other. But that didnt work. 

Also, I rebased the changes on top of master and later pulled from this branch which caused some issues (started showing 26 files as changed), hence I squashed the commits and after ensuring it is rebased on top of latest master, have force pushed it. Sorry if this makes the review difficult

raghav710:
@HowardWolosky  these tests require the Contents API to be implemented. Have created an issue for that and will be working on it

raghav710:
@HowardWolosky this is now ready for review. Please let me know your comments on this.

GitHubReferences.ps1
HowardWolosky: Looks like it would now be [List Matching References](https://developer.github.com/v3/git/refs/#list-matching-references).

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+
+        [Parameter(ParameterSetName='Elements')]
+        [string] $RepositoryName,
+
+        [Parameter(
+            Mandatory,
+            ParameterSetName='Uri')]
+        [string] $Uri,
+
+        [string] $Reference,
```
HowardWolosky, line 64: As I've thought about this more, I don't think it makes sense for users to have to know the syntax `heads/<branch name>` or `tags/<tag name>`.  Instead, I think you should split this out and have an optional parameter of `[string] $Tag` and another one as `[string] $Head`.  Users can just specify the one that should be used. You can defined them to be mutually exclusive via parametersets.

GitHubReferences.ps1
HowardWolosky: Some thought needs to go in to how you would make a combined single/multiple function given the additional suggestion of splitting out `$Reference` as discussed below.

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+        'ProvidedReference' = $PSBoundParameters.ContainsKey('Reference')
+    }
+
+    if ($OwnerName -xor $RepositoryName)
+    {
+        $message = 'You must specify both Owner Name and Repository Name.'
```
HowardWolosky, line 85: That's not what `-xor` does.  Your condition, as written right now means:
"Either OwnerName or RepositoryName must have a value, but BOTH can't have a value."
So, either your condition is wrong or your error message is wrong.

That being said, you don't need to verify this at all.  By default, `Resolve-RepositoryElements` will throw an exception if it isn't returning with a non-empty OwnerName and RepositoryName. Linking to the documentation: https://github.com/microsoft/PowerShellForGitHub/blob/6735ba57a5a43b61a37ef09d4021296dcd417dba/GitHubCore.ps1#L784-L788

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+        Write-Log -Message $message -Level Error
+        throw $message
+    }
+
+    if (-not [String]::IsNullOrEmpty($RepositoryName))
```
HowardWolosky, line 90: We'll never be in a situation where we don't have a repo name, per the above.

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+    {
+        $uriFragment = "repos/$OwnerName/$RepositoryName/git/matching-refs"
```
HowardWolosky, line 92: Where are you handling the [singular reference](https://developer.github.com/v3/git/refs/#get-a-single-reference) case?

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+        $description =  "Getting all references for $RepositoryName"
+        if ($PSBoundParameters.ContainsKey('Reference'))
+        {
+            $uriFragment = $uriFragment + "/$Reference"
+            $description =  "Getting Reference $Reference for $RepositoryName"
+        }
+        else
+        {
+            # Add a slash at the end. Calling it with only matching-refs without the slash causes a 404
+            $uriFragment = $uriFragment + "//"
```
HowardWolosky, line 102: If this is true, then the logic above is a bit delicate.  What if `$Reference` has a leading slash?  What happens then?  I think you need to do some more robust input validation.

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+        [Parameter(ParameterSetName='Elements')]
+        [string] $RepositoryName,
+
+        [Parameter(
+            Mandatory,
+            ParameterSetName='Uri')]
+        [string] $Uri,
+
+        [Parameter(Mandatory)]
+        [string] $Reference,
```
HowardWolosky, line 181: Same feedback as for `Get-*`.  Instead of making users understand the `head/<branch name>` / `tag/<tag name>` syntax, why not just provide explicit parameters that can be used?

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+    $elements = Resolve-RepositoryElements -BoundParameters $PSBoundParameters -DisableValidation
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    if ($OwnerName -xor $RepositoryName)
```
HowardWolosky, line 202: Same feedback as above.

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
```
HowardWolosky, line 316: Same feedback as above.

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+        [Parameter(ParameterSetName='Elements')]
+        [string] $RepositoryName,
+
+        [Parameter(
+            Mandatory,
+            ParameterSetName='Uri')]
+        [string] $Uri,
+
+        [Parameter(Mandatory)]
+        [string] $Reference,
```
HowardWolosky, line 293: Same feedback as above.

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
```
HowardWolosky, line 402: Same feedback as above.

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+    $elements = Resolve-RepositoryElements -BoundParameters $PSBoundParameters -DisableValidation
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    if ($OwnerName -xor $RepositoryName)
```
HowardWolosky, line 420: Same feedback as above.

GitHubReferences.ps1
```diff
@@ -0,0 +1,442 @@
...
+    {
+        $message = 'You must specify both Owner Name and Repository Name.'
+        Write-Log -Message $message -Level Error
+        throw $message
+    }
+
+    $uriFragment = "repos/$OwnerName/$RepositoryName/git/refs/$Reference"
+    $description =  "Deleting Reference $Reference from repository $RepositoryName"
+
+    $params = @{
```
HowardWolosky, line 430: A recent PR introduced the idea of getting user confirmation on delete actions (which can be overridden by a commandline parameter).  I think it's a pretty solid idea, and I think we should apply that approach on all new "delete" methods.   Take a [look here](https://github.com/microsoft/PowerShellForGitHub/commit/536b762425d51a181166c2c47ad2b00014911d1d#diff-2a0b254e60269a3a0794d63c7922522dR518-R549) for more info.

I'd ask that you _also_ add an additional Example to here that shows how to avoid the prompt (as that reference that I pointed to does as well).

raghav710:
Thanks for the comments @HowardWolosky  I'll take a look at the changes requested and try to get back as soon as possible. Sorry it has taken a year for this since I wasn't continuously involved

GitHubReferences.ps1
raghav710: Agreed. That looks like a good idea. 
suggestion: Can we have the flag as -NoConfirm or similar so that the user need not type in -Confirm:$false ?

GitHubReferences.ps1
raghav710: I picked up this check from [GitHubIssues.ps1](https://github.com/microsoft/PowerShellForGitHub/blob/master/GitHubIssues.ps1#L195) where there is a validation check to ensure that both OwnerName and RepositoryName are provided and would throw an error message (as is done here) when only one of them is provided (hence the xor)

I agree on doing away with this validation as Resolve-RepositoryElements takes care of it anyway!

GitHubReferences.ps1
raghav710: The additional slash at the end is required because Invoke-GHRestMethod removes trailing slashes.
https://github.com/microsoft/PowerShellForGitHub/blob/d279f1c7e1bd8261cdcf9dab6a8b78d3a0004457/GitHubCore.ps1#L126-L130
In this case, since there would not be any other variable appended, no input validation is required
And for the part where we do
`$uriFragment = $uriFragment + "/$Reference"`

Validation would not be required since the $reference variable would now be generated from the Tag/Branch passed in by the user

Please feel free to correct me if I misunderstood this comment

HowardWolosky:
@raghav710 -- can you confirm that this is ready for re-review?

Additionally, I used GitHub's merge conflict resolver for you on the manifest file, and it appears that it messed up the line breaks in the file, causing the whole file to appear to have been changed.  Can you please push two new commits, one where you revert the file to its original state, and the second where you re-apply your specific changes, so that only the changes are seen in the diff?  Thanks!

raghav710:
@HowardWolosky sure. I'll do that. I checked the tests and i remember them running fine. Will check the code once again for any mistakes and planning to drop a note max by tomorrow that this is ready for review. Sorry this has taken so long

GitHubReferences.ps1
raghav710: Updated this in the latest commit

GitHubReferences.ps1
raghav710: This is now done

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+    .PARAMETER Tag
+        The name of the Tag to be retrieved.
+
+    .PARAMETER Branch
+        The name of the Branch to be retrieved.
+
+    .PARAMETER All
+        If provided, will retrieve both branches and tags for the given repository
+
+    .PARAMETER MatchPrefix
```
raghav710, line 36: This is the option for calling the matching-refs API. Wasn't sure what is the standard pattern for such a functionality

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+    .EXAMPLE
+        Get-GitHubReference -OwnerName microsoft -RepositoryName PowerShellForGitHub -Branch master
+
+    .EXAMPLE
+        Get-GitHubReference -OwnerName microsoft -RepositoryName PowerShellForGitHub -Branch powershell -MatchPrefix
+
+        Get the branch 'powershell' and if it doesn't exist, get all branches beginning with 'powershell'
+
+    .EXAMPLE
+        Get-GitHubReference -Uri https://github.com/You/YourRepo -All
```
raghav710, line 65: I had to add an 'All' parameter because of the parametersets. Since both branch and tag are optional, calling Get-GithubReference without parameters results in PS not being able to find a parameterset, hence this workaround. Appreciate any suggestions on bettering this

raghav710:
@HowardWolosky this is now ready for another round of review. Sorry there are multiple commits, I would like to squash these before merging.

Please let me know your thoughts on the Get-GithubReference, I had to take a few decisions with the parameters to accomodate the different use cases. Have also updated the tests to be more extensive.

raghav710:
@HowardWolosky please feel free to comment on the approach used here and whether it is good enough. Also if you think some of this functionality is already being added by #200 I'm fine with removing parts of these changes so we can bring in only what is required

raghav710:
Thanks for the comments @HowardWolosky . I would take some time in getting back to this. Is there a timeline we are looking at to get this merged? (So the dependent PRs dont have to wait?) and I can try working with that in mind

HowardWolosky:
> Thanks for the comments @HowardWolosky Howard Wolosky FTE . I would take some time in getting back to this. Is there a timeline we are looking at to get this merged? (So the dependent PRs dont have to wait?) and I can try working with that in mind

Do you think you can get to it within the next two weeks?  If not, then maybe someone else in the community might be able to apply the feedback to what you've done so far, in order to unblock the other PR's waiting on this (#200 is definitely waiting on this, but there are others like #193 which need it for their UT's).

You've put a bunch of effort into this so far, and I want to ensure that you're getting the appropriate credit for your work.  I also want to keep the momentum going for the others that are also actively contributing to the project and prevent blockages from happening for too long.

Thanks for your understanding and your participation!

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+            'TelemetryProperties' = $telemetryProperties
+            'NoStatus' = (Resolve-ParameterWithDefaultConfigurationValue -Name NoStatus -ConfigValueName DefaultNoStatus)
+        }
+
+        return Invoke-GHRestMethod @params
+    }
+
+}
+
+function Resolve-Reference($Tag, $Branch)
```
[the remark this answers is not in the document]
raghav710: @HowardWolosky does this apply to helper functions which we'd not want to expose directly to the user as well. Is there a different convention I should use for such functions?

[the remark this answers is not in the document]
HowardWolosky: It applies to _all_ functions, even those not exported.  The CBH can be just as helpful for the maintainer as it can be for the end-user.

raghav710:
@HowardWolosky I plan to work on this today and fix the code based on comments

GitHubReferences.ps1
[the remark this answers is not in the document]
raghav710: Agreed. Will update this one

GitHubReferences.ps1
raghav710: Thanks. I think this greatly simplifies it!

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+        Get-GitHubReference -Uri https://github.com/You/YourRepo -All
+
+        Get all references in the repository specified by URI
+#>
+    [CmdletBinding(
+        SupportsShouldProcess,
+        DefaultParameterSetName='Elements')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "", Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
+    param(
+        [Parameter(Mandatory, ParameterSetName='BranchElements')]
```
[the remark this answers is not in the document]
raghav710: Will update this

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+    $elements = Resolve-RepositoryElements -BoundParameters $PSBoundParameters -DisableValidation
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    $reference = Resolve-Reference $Tag $Branch
```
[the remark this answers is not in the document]
raghav710: Will fix this as well as ensuring there is CBH for the method

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+        'AccessToken' = $AccessToken
+        'TelemetryEventName' = $MyInvocation.MyCommand.Name
+        'TelemetryProperties' = $telemetryProperties
+        'NoStatus' = (Resolve-ParameterWithDefaultConfigurationValue -Name NoStatus -ConfigValueName DefaultNoStatus)
+    }
+
+    return Invoke-GHRestMethod @params
+}
+
+function Update-GitHubReference
```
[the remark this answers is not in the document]
raghav710: Fixed in the next commit

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+
+    .EXAMPLE
+        Remove-GitHubReference -Uri https://github.com/You/YourRepo -Branch master
+
+    .EXAMPLE
+        Remove-GitHubReference -OwnerName microsoft -RepositoryName PowerShellForGitHub -Tag milestone1 -Confirm:$false
+
+        Remove the tag milestone1 without prompting for confirmation
+    #>
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact="High")]
```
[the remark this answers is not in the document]
raghav710: Fixed in the next commit

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+        [string] $Branch,
+
+        [string] $AccessToken,
+
+        [switch] $NoStatus
+    )
+
+    $repositoryInfoForDisplayMessage = if ($PSCmdlet.ParameterSetName -eq "Uri") { $Uri } else { $OwnerName, $RepositoryName -join "/" }
+    $reference = Resolve-Reference $Tag $Branch
+    if ($PSCmdlet.ShouldProcess($repositoryInfoForDisplayMessage, "Remove reference: $reference"))
```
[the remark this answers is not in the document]
raghav710: Fixed in next commit

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+            'Description' = $description
+            'AccessToken' = $AccessToken
+            'TelemetryEventName' = $MyInvocation.MyCommand.Name
+            'TelemetryProperties' = $telemetryProperties
+            'NoStatus' = (Resolve-ParameterWithDefaultConfigurationValue -Name NoStatus -ConfigValueName DefaultNoStatus)
+        }
+
+        return Invoke-GHRestMethod @params
+    }
+
```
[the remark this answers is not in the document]
raghav710: Fixed in next commit

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+        [switch] $MatchPrefix,
+
+        [string] $AccessToken,
+
+        [switch] $NoStatus
+    )
+
+    Write-InvocationLog -Invocation $MyInvocation
+
+    $elements = Resolve-RepositoryElements -BoundParameters $PSBoundParameters -DisableValidation
```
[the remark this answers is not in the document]
raghav710: Thanks for the note. As mentioned in #243 this is one of the copy-paste issues :) will fix this

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+        'ProvidedReference' = $PSBoundParameters.ContainsKey('Reference')
```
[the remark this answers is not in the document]
raghav710: Removed this. was an artifact of the previous code where branches/tags were not parameters

GitHubReferences.ps1
```diff
@@ -0,0 +1,525 @@
...
+    }
+
+    $uriFragment = "repos/$OwnerName/$RepositoryName/git"
+
+    if ($All) {
+        # Add a slash at the end as Invoke-GHRestMethod removes the last trailing slash. Calling this API without the slash causes a 404
```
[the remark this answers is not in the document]
raghav710: I am open for any suggestions to making this better. At this point, this is all I'm able to come up with :(

Tests/GitHubReferences.Tests.ps1
```diff
@@ -0,0 +1,466 @@
...
+<#
+.Synopsis
+   Tests for GitHubReferences.ps1 module
+#>
+
+# This is common test code setup logic for all Pester test files
+$moduleRootPath = Split-Path -Path $PSScriptRoot -Parent
+. (Join-Path -Path $moduleRootPath -ChildPath 'Tests\Common.ps1')
+
+function Update-MasterSHA()
```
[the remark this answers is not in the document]
raghav710: Updated to use Set-Content

raghav710:
@HowardWolosky I have made an attempt at adding pipeline support and refactored where required to match the current standards. Also updated the tests to check for pipelines and to follow Pester standards where applicable.

Taking this forward, would someone in the community be willing to take this from here? Given that there are other PRs waiting on this, I dont want to delay this any longer waiting for my availability.

PS: its amazing how much has been done in the project in a short period! from pipeline support to removing boilerplate from tests to standardizing many code conventions. Kudos to you and the community!

HowardWolosky:
Thanks for all your efforts here, @raghav710.  You've provided a solid base for the remaining work that needs to be done.  I'm nearing completion on an update to this work to fill in the remaining gaps.  Once I'm done with that, I'll close out this PR and start a new one.  Again, thanks so much.  Looking forward to seeing your further contributions in this project when you have time.

raghav710:
Thanks @HowardWolosky for giving me a chance to contribute to this project. Again, apologies for not taking this to completion

PR #100, opened 2019-04-03
Rename ConvertFrom-Markdown

### LINKED ISSUE #99: Rename ConvertFrom-Markdown
`ConvertFrom-Markdown` was added as a standard command in [PowerShellCore 6](https://github.com/PowerShell/PowerShell/commit/f8b3abaac7541d4ebf294ab8142222d06f65da6a).

Given that, this module's command should be renamed in order to avoid any conflicts when used in that environment.

Current proposal is to rename it to `ConvertFrom-GitHubMarkdown` -- adding `GitHub` in there is more consistent with the rest of the module commands.  The only downside to this name is that it might imply that it's specifically about converting _GitHub Flavored_ Markdown whereas it [converts both "standard" and "GitHub flavored"](https://github.com/PowerShell/PowerShellForGitHub/blob/68fa5a563b96490774e9c5fae38193194f765157/GitHubMiscellaneous.ps1#L130)

HowardWolosky:
PowerShellCore 6 added a `ConvertFrom-Markdown` command, and this module's
command therefore causes a conflict.

Renaming to `ConvertFrom-GitHubMarkdown` in an effort to best identify this
as being a "GitHub" method that converts Markdown, although it unfortunately
does imply that it might only support "GitHub Flavored Markdown", where in fact
it supports both that as well as standard.

Resolves #99 - Rename ConvertFrom-Markdown

### DIFF
--- a/GitHubMiscellaneous.ps1
+++ b/GitHubMiscellaneous.ps1
@@ -73,7 +73,7 @@ function Get-GitHubRateLimit
     return Invoke-GHRestMethod @params
 }
 
-function ConvertFrom-Markdown
+function ConvertFrom-GitHubMarkdown
 {
 <#
     .SYNOPSIS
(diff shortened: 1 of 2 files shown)


HowardWolosky:
Will leave this up for a couple days for comment to see if anyone cares about this name change or has alternative suggestions, before I proceed with merging it in.

cc: @cliffchapmanrbx

PR #101, opened 2019-04-08
Add URL config for GitHub Enterprise

### LINKED ISSUE #98: Support GitHub Enterprise
GitHub Enterprise supports nearly the same API for interacting, and it appears relatively straightforward to provide a way to override the otherwise hardcoded API URLs in `GitHubCore.ps1` to local GHE server environments.

Adding GHE support appears to be low hanging fruit. Is there any prior work done on this? Any pitfalls?

Cellivar:
I think I ran the pester tests correctly, or at least they all came back green for me anyhow. Static analysis also came back empty.

This resolves #98 with the ability to supply a new configuration parameter, `GitHubBaseUrl`. It will default to `github.com`. GitHub Enterprise installs to a different URL and the API remains as `api.somegithub.com`, thus it's simple to hit that API on a different base URL.

I've been using a similar set of changes along with a few other quality of life ones in my organization over the past week as we prototyped out some tools around automating GitHub Enterprise. It performed admirably with these changes here.

Fixed a related typo in the documentation as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -248,13 +248,15 @@ Here are some general guidelines
 
 If you want to add a new configuration value to the module, you must modify the following:
  * In `Import-GitHubConfiguration`, update `$config` to declare the new property along with
-   it's default value, being sure that PowerShell will understand what its type is.
+   its default value, being sure that PowerShell will understand what its type is. Properties
+   should be alphabetical.
  * Update `Get-GitHubConfiguration` and add the new property name to the `ValidateSet` list
    so that tab-completion and documentation gets auto-updated. You shouldn't have to add anything
-   to the body of the method.
+   to the body of the method. Property names should be alphabetical.
  * Add a new explicit parameter to `Set-GitHubConfiguration` to receive the property, along with
    updating the CBH (Comment Based Help) by adding a new `.PAREMETER` entry. You shouldn't
-   have to add anything to the body of the method.
+   have to add anything to the body of the method. Parameters should be alphabetical save for the
+   `SessionOnly` switch, which should be last.
 
 ----------
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -247,7 +247,7 @@ Here are some general guidelines
 ## Adding New Configuration Properties
 
 If you want to add a new configuration value to the module, you must modify the following:
- * In `Restore-GitHubConfiguration`, update `$config` to declare the new property along with
+ * In `Import-GitHubConfiguration`, update `$config` to declare the new property along with
    it's default value, being sure that PowerShell will understand what its type is.
  * Update `Get-GitHubConfiguration` and add the new property name to the `ValidateSet` list
    so that tab-completion and documentation gets auto-updated. You shouldn't have to add anything
--- a/GitHubConfiguration.ps1
+++ b/GitHubConfiguration.ps1
@@ -253,7 +253,7 @@ function Get-GitHubConfiguration
 
         Always returns the value for this session, which may or may not be the persisted
         setting (that all depends on whether or not the setting was previously modified
-        during this session using Set-GitHubCOnfiguration -SessionOnly).
+        during this session using Set-GitHubConfiguration -SessionOnly).
 
         The Git repo for this module can be found here: http://aka.ms/PowerShellForGitHub
 
--- a/GitHubConfiguration.ps1
+++ b/GitHubConfiguration.ps1
@@ -67,6 +67,11 @@ function Set-GitHubConfiguration
 
         The Git repo for this module can be found here: http://aka.ms/PowerShellForGitHub
 
+    .PARAMETER ApiHostName
+        The hostname of the GitHub instance to communicate with. Defaults to 'github.com'. Provide a
+        different hostname when using a GitHub Enterprise server. Do not include the HTTP/S prefix,
+        and do not include 'api'. For example, use "github.contoso.com".
+
     .PARAMETER ApplicationInsightsKey
         Change the Application Insights instance that telemetry will be reported to (if telemetry
         hasn't been disabled via DisableTelemetry).
@@ -146,11 +151,6 @@ function Set-GitHubConfiguration
         created/updated and the specified configuration changes will only remain in memory/effect
         for the duration of this PowerShell session.
 
-    .PARAMETER GitHubBaseUrl
-        The base URL of the GitHub instance to communicate with. Defaults to 'github.com'. Provide
-        a different URL when using a GitHub Enterprise server. The server must respond to the
-        standard API interface, for instance, api.github.contoso.com.
-
     .EXAMPLE
         Set-GitHubConfiguration -WebRequestTimeoutSec 120 -SuppressNoTokenWarning
 
@@ -165,14 +165,17 @@ function Set-GitHubConfiguration
         session only.
 
     .EXAMPLE
-        Set-GitHubConfiguration -GitHubBaseUrl "github.contoso.com"
+        Set-GitHubConfiguration -ApiHostName "github.contoso.com"
 
         Sets all requests to connect to a GitHub Enterprise server running at
-        api.github.contoso.com.
+        github.contoso.com.
 #>
     [CmdletBinding(SupportsShouldProcess)]
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSShouldProcess", "", Justification="Methods called within here make use of PSShouldProcess, and the switch is passed on to them inherently.")]
     param(
+        [ValidatePattern('^(?!https?:)(?!api\.)(?!www\.).*')]
+        [string] $ApiHostName,
+
         [string] $ApplicationInsightsKey,
 
         [string] $AssemblyPath,
@@ -208,9 +211,7 @@ function Set-GitHubConfiguration
         [ValidateRange(0, 3600)]
         [int] $WebRequestTimeoutSec,
 
-        [switch] $SessionOnly,
-
-        [string] $GitHubBaseUrl
+        [switch] $SessionOnly
     )
 
     $persistedConfig = $null
@@ -270,6 +271,7 @@ function Get-GitHubConfiguration
     param(
         [Parameter(Mandatory)]
         [ValidateSet(
+            'ApiHostName',
             'ApplicationInsightsKey',
             'AssemblyPath',
             'DefaultNoStatus',
@@ -286,8 +288,7 @@ function Get-GitHubConfiguration
             'RetryDelaySeconds',
             'SuppressNoTokenWarning',
             'SuppressTelemetryReminder',
-            'WebRequestTimeoutSec',
-            'GitHubBaseUrl')]
+            'WebRequestTimeoutSec')]
         [string] $Name
     )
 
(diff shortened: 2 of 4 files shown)


Cellivar:
AppVeyor CI is not visible for me so I can't check :(

GitHubCore.ps1
```diff
@@ -154,7 +151,9 @@ function Invoke-GHRestMethod
     # we'll just always continue the existing one...
     $stopwatch.Start()
 
-    $url = "$script:gitHubApiUrl/$UriFragment"
+    $gitHubApiUrl = "https://api.$(Get-GitHubConfiguration -Name "GitHubBaseUrl")"
```
HowardWolosky: I'd propose a different breakdown for readability of components. (Can't use GitHub's suggestion feature since it only supports single-line):

```
$gitHubBaseUrl = Get-GitHubConfiguration -Name "GitHubBaseUrl"
$url = "https://api.$gitHubBaseUrl/$UriFragment"
```

or, you could do this:

```
$gitHubBaseUrl = Get-GitHubConfiguration -Name "GitHubBaseUrl"
$gitHubApiUrl = "https://api.$($gitHubBaseUrl)"
$url = "$gitHubApiUrl/$UriFragment"
```

While what you have already is totally valid, I think embedding function calls within a string definition reduces readability of the code.

GitHubCore.ps1
```diff
@@ -735,8 +734,10 @@ function Split-GitHubUri
         repositoryName = [String]::Empty
     }
 
-    if (($Uri -match '^https?://(?:www.)?github.com/([^/]+)/?([^/]+)?(?:/.*)?$') -or
-        ($Uri -match '^https?://api.github.com/repos/([^/]+)/?([^/]+)?(?:/.*)?$'))
+    $gitHubUrlBase = $(Get-GitHubConfiguration -Name "GitHubBaseUrl")
```
HowardWolosky: I'd suggest being consistent with how this is referred to (`UrlBase` vs `BaseUrl`):

```suggestion
    $gitHubBaseUrl = $(Get-GitHubConfiguration -Name "GitHubBaseUrl")
```

GitHubCore.ps1
```diff
@@ -735,8 +734,10 @@ function Split-GitHubUri
...
+
+    if (($Uri -match "^https?://(?:www.)?$gitHubUrlBase/([^/]+)/?([^/]+)?(?:/.*)?$") -or
```
```suggestion
    if (($Uri -match "^https?://(?:www.)?$gitHubBaseUrl/([^/]+)/?([^/]+)?(?:/.*)?$") -or
```

GitHubCore.ps1
```diff
@@ -735,8 +734,10 @@ function Split-GitHubUri
...
+        ($Uri -match "^https?://api.$gitHubUrlBase/repos/([^/]+)/?([^/]+)?(?:/.*)?$"))
```
```suggestion
        ($Uri -match "^https?://api.$gitHubBaseUrl/repos/([^/]+)/?([^/]+)?(?:/.*)?$"))
```

GitHubConfiguration.ps1
```diff
@@ -197,7 +208,9 @@ function Set-GitHubConfiguration
         [ValidateRange(0, 3600)]
         [int] $WebRequestTimeoutSec,
 
-        [switch] $SessionOnly
+        [switch] $SessionOnly,
+
+        [string] $GitHubBaseUrl
```
HowardWolosky: Would be worthwhile to help prevent users from shooting themselves in the foot by validating what they're passing in and ensuring that they're not including `http://`, `https://`, `www.` or `api.` as part of their BaseUrl:

```
   [ValidatePattern('^(?!https?)(?!api\.)(?!www\.).*')]
```

GitHubConfiguration.ps1
```diff
@@ -146,6 +146,11 @@ function Set-GitHubConfiguration
         created/updated and the specified configuration changes will only remain in memory/effect
         for the duration of this PowerShell session.
 
+    .PARAMETER GitHubBaseUrl
```
HowardWolosky: In terms of naming for this configuration, I think I have two input suggestions here:
1. Instead of saying `GitHub` as part of this name, I think it should be `Api`.
2. This doesn't seem to be a `Url`, but rather a `HostName`.

So, I think it's probably more correct to call this the `ApiHostName`.

GitHubConfiguration.ps1
```diff
@@ -146,6 +146,11 @@ function Set-GitHubConfiguration
...
+        The base URL of the GitHub instance to communicate with. Defaults to 'github.com'. Provide
```
HowardWolosky: I think the description here should be updated to explicitly indicate that users should leave off `api.` from the host name in the example that you provided.

GitHubConfiguration.ps1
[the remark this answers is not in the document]
Cellivar: I can. Historically when I have added functionality to systems, preserving backwards compatibility trumped formatting. In this case moving the parameter up in the parameter list would break compatibility in the event that the function is called without explicit argument names (`Test-Path "/some/dir"` instead of `Test-Path -Path "/some/dir"`.)

If you're okay with breaking backwards compatibility I'm fine with reordering it to alphabetical.

GitHubCore.ps1
Cellivar: The readability concern is quite valid, I was going for minimal changes in my first iteration. I'll modify according to your recommendation.

GitHubConfiguration.ps1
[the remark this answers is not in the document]
HowardWolosky: A concern on back compat is _quite_ valid.  For this method _specifically_ though, I'm not concerned.  The usage pattern for this method is such that it is _highly_ unlikely that users would ever call this and specify _all_ of the parameters simply by position.  Users are much more likely to either call it multiple times specifying one parameter each time, or to call it a single time and explicitly specify the parameter name with its value.

So, while I _completely_ agree that we should avoid back compat breaking changes to parameter ordering _in general_, I'm _not_ concerned with that problem with _this_ function specifically.

Thanks for raising the concern though.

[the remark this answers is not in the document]
Cellivar: Excellent, I've got some time before bed so expect an updated changeset for review shortly.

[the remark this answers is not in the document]
Cellivar: Reordered according to naming.

GitHubConfiguration.ps1
```diff
@@ -273,7 +286,8 @@ function Get-GitHubConfiguration
             'RetryDelaySeconds',
             'SuppressNoTokenWarning',
             'SuppressTelemetryReminder',
-            'WebRequestTimeoutSec')]
+            'WebRequestTimeoutSec',
+            'GitHubBaseUrl')]
```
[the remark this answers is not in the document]
Cellivar: Reordered according to naming.

GitHubConfiguration.ps1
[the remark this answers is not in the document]
Cellivar: Reordered according to naming.

GitHubConfiguration.ps1
```diff
@@ -606,6 +620,7 @@ function Import-GitHubConfiguration
         'suppressNoTokenWarning' = $false
         'suppressTelemetryReminder' = $false
         'webRequestTimeoutSec' = 0
+        'gitHubBaseUrl' = 'github.com'
```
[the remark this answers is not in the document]
Cellivar: Reordered according to naming.

GitHubCore.ps1
Cellivar: Reordered according to first suggestion, as I quite liked how the syntax highlighting looked and the statement reads from left to right.

GitHubCore.ps1
Cellivar: Resolved with a change to 'hostName' a la API host name.

GitHubConfiguration.ps1
Cellivar: Suggestion implemented, I like it better 👍

GitHubConfiguration.ps1
Cellivar: Clarified do-nots and a clear example.

GitHubConfiguration.ps1
Cellivar: Implemented with a minor addition, I've added a colon to the end of the first negative lookahead so that `https://git.contoso.com` would be caught while `httpgithub.contoso.com` would be fine. 

This uh, this one comes from experience with an _interesting_ company. 🙃

GitHubCore.ps1
```diff
@@ -154,7 +151,9 @@ function Invoke-GHRestMethod
     # we'll just always continue the existing one...
     $stopwatch.Start()
 
-    $url = "$script:gitHubApiUrl/$UriFragment"
+    $hostName = $(Get-GitHubConfiguration -Name "ApiHostName")
+
+    $url = "https://api.$hostname/$UriFragment"
```
HowardWolosky: nit: Can you use the same casing as the variable declaration?

```suggestion
    $url = "https://api.$hostName/$UriFragment"
```

GitHubCore.ps1
```diff
@@ -735,8 +734,10 @@ function Split-GitHubUri
         repositoryName = [String]::Empty
     }
 
-    if (($Uri -match '^https?://(?:www.)?github.com/([^/]+)/?([^/]+)?(?:/.*)?$') -or
-        ($Uri -match '^https?://api.github.com/repos/([^/]+)/?([^/]+)?(?:/.*)?$'))
+    $hostName = $(Get-GitHubConfiguration -Name "ApiHostName")
```
HowardWolosky: It's fine to keep this as is, but the preceeding `$` is unnecessary here since we're not within a double-quoted string.  Technically the parenthesis are unnecessary too, but are totally acceptable for clarity in expression grouping.  Just calling this out for your awareness.

CONTRIBUTING.md
```diff
@@ -247,7 +247,7 @@ Here are some general guidelines
 ## Adding New Configuration Properties
 
 If you want to add a new configuration value to the module, you must modify the following:
- * In `Restore-GitHubConfiguration`, update `$config` to declare the new property along with
+ * In `Import-GitHubConfiguration`, update `$config` to declare the new property along with
```
HowardWolosky: I think you might have missed my previous PR comment about this....so adding it back here for your attention:

I think it would _also_ be great to include a minor documentation update to point out this specific configuration.  It probably makes sense to call it out within the `Configuration` section of the README.

GitHubCore.ps1
Cellivar: Yep, I waffled a bit on it in light of your code readability feedback and for giggles have spent the last couple of days keeping sub-commands inside nested parenthesis like this to see how the effect grows with more code around it. I'm still undecided on keeping parenthesis or leaving them off so I'll defer to your choice since you're dealing with this code more often than I.

GitHubCore.ps1
Cellivar: This was in fact a typo, nice catch 👀

CONTRIBUTING.md
Cellivar: I did miss that, my apologies. Still getting used to the GitHub PR interface, I'm more used to the more compact Azure Repos comment boxes and I think I got the two comments mixed up.

HowardWolosky: No worries.  Thanks for adding it, as well as the updates to contrbuting re: keeping those parameters alphabetical.

Cellivar:
No problem! Thank you for your patient guidance while I shook the rust off my public contribution skills.

PR #118, opened 2019-07-02
Fix hostname support for Enterprise installations of GitHub

### LINKED ISSUE #117: Fix api url for Github Enterprise
The api url has been coded in the following format: 
```powershell
$url = "https://api.$hostName/$UriFragment"
```

While this works for github it fails with the enterprise version because the api url is different:
```code
http(s)://[hostname]/api/v3
```

As a result when you call any of the function referencing Github enterprise you may get the following error:

```powershell
The remote name could not be resolved: 'api.myghhost.com'
At C:\Program 
Files\WindowsPowerShell\Modules\PowerShellForGitHub\0.8.0\GitHubCore.ps1:520 
char:9
+         throw $newLineOutput
+         ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (The remote name.... 
   com':String) [], RuntimeException
    + FullyQualifiedErrorId : The remote name could not be resolved: 'api.myghhost.com'

```

Will raise a PR to fix this.

v2kiran:
Per the [documentation](https://developer.github.com/enterprise/2.16/v3/#current-version), enterprise installations support a base URL of  `http(s)://[hostname]/api/v3` in addition to the existing support of `http(s)://api.[hostname]/`.  This PR updates the module to _always_ use the `http(s)://[hostname]/api/v3` syntax when `[hostname]` is _not_ `github.com`.

Fixes #117

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "editor.formatOnSave": false,
+    "editor.formatOnType": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    "editor.formatOnSave": false,
-    "editor.formatOnType": false
-}
--- a/GitHubCore.ps1
+++ b/GitHubCore.ps1
@@ -2,20 +2,20 @@
 # Licensed under the MIT License.
 
 @{
-    defaultAcceptHeader = 'application/vnd.github.v3+json'
-    mediaTypeVersion = 'v3'
+    defaultAcceptHeader  = 'application/vnd.github.v3+json'
+    mediaTypeVersion     = 'v3'
     squirrelAcceptHeader = 'application/vnd.github.squirrel-girl-preview'
     symmetraAcceptHeader = 'application/vnd.github.symmetra-preview+json'
 
- }.GetEnumerator() | ForEach-Object {
-     Set-Variable -Scope Script -Option ReadOnly -Name $_.Key -Value $_.Value
- }
+}.GetEnumerator() | ForEach-Object {
+    Set-Variable -Scope Script -Option ReadOnly -Name $_.Key -Value $_.Value
+}
 
 Set-Variable -Scope Script -Option ReadOnly -Name ValidBodyContainingRequestMethods -Value ('Post', 'Patch', 'Put', 'Delete')
 
 function Invoke-GHRestMethod
 {
-<#
+    <#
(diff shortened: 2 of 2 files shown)


GitTill:
remarks regarding this code change:

GitLab API Version (and url) is updated to V4 -> https://docs.gitlab.com/ee/api/v3_to_v4.html. This change is still using V3 (not supported in GitLab 11.0 and higher). Just changing the URL to v4 results in authentication errors in my environment (tested with valid token): 

> 404 Not Found
    This typically happens when the current user isn't properly authenticated.  You may need an Access Token with additional scopes checked.

This change is only supporting https - but not http. Support for http would be great.

HowardWolosky:
>remarks regarding this code change:
>
>GitLab API Version (and url) is updated to V4
>... 

@GitTill - The GitLab and GitHub API's are quite different. This module is not designed to work with GitLab...only GitHub.

I took a look at their website and found a reference to [PSGitlab](https://github.com/ngetchell/PSGitLab) is an unofficial PowerShell client for their API. You may want to look at that module.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
HowardWolosky: At present time, I don't want to check-in editor-specific files into the project.  I realize that you added this as a result of needing to undo your other formatting changes , and I'm open to being convinced otherwise, but at the moment I'm not.

GitHubCore.ps1
```diff
@@ -153,7 +153,14 @@ function Invoke-GHRestMethod
 
     $hostName = $(Get-GitHubConfiguration -Name "ApiHostName")
 
-    $url = "https://api.$hostName/$UriFragment"
+    if ($hostName -eq "github.com")
```
HowardWolosky: I have generally been avoiding double-quoted strings unless I am expecting for them to be interpreted due to embedded expressions/variables/etc...

```suggestion
    if ($hostName -eq 'github.com')
```

v2kiran: :) sure...i will change to single

.vscode/settings.json
v2kiran: ah ...how do i remove the file...just add it to gitignore?

HowardWolosky: 1. delete the file
2. `git add .vscode/settings.json`
3. `git commit`
4. `git push`

v2kiran: thanks...this is done

PR #143, opened 2020-01-09
Adding Get Repository.Contents functionality

Shazwazza:
I'm unsure of if you will accept PRs that only implement one part of the functionality. For example, for the Repository.Contents there is: Get, Create/Update, Delete, Get archive link

But I'm really only looking to implement the `Get` part. Is that ok for a phase 1 approach?

Please let me know what else needs to be considered, i haven't looked at the tests yet so I'm unsure if everything is tested or not.


Shazwazza:
I've marked this 'ready' for review since I have this working locally with this code for retrieving file contents, you can pass in -Path to get a single item or leave it out to get the root.

GitHubContents.ps1
```diff
@@ -0,0 +1,93 @@
+function Get-GitHubContents
```
HowardWolosky: You also need to make an update to `PowerShellForGitHub.psd1` and add this file to `NestedModules` and the function to `FunctionsToExport` (both alphabetically please).

GitHubContents.ps1
```diff
@@ -0,0 +1,93 @@
...
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+        'Path' = $PSBoundParameters.ContainsKey('Path')
+    }
+
+    $uriFragment = [String]::Empty
+    $description = [String]::Empty    
+
+    $uriFragment = "/repos/$OwnerName/$RepositoryName/contents"
```
HowardWolosky: nit: Please remove trailing spaces

GitHubContents.ps1
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    $uriFragment = [String]::Empty
+    $description = [String]::Empty    
+
+    $uriFragment = "/repos/$OwnerName/$RepositoryName/contents"    
+
+    if ($PSBoundParameters.ContainsKey('Path'))
+    {
+        $uriFragment += "/$Path"
```
HowardWolosky: You may need to verify that `$Path` doesn't start with a `/` or `\` (and if it does, trim it first)

GitHubContents.ps1
```diff
@@ -0,0 +1,93 @@
...
+    }
+    else 
+    {
+        $description = "Getting all content for in $RepositoryName"
+    }
+
+    $params = @{
+        'UriFragment' = $uriFragment
+        'Description' =  $description
+        'AcceptHeader' = (Get-MediaAcceptHeader -MediaType $MediaType -AcceptHeader $symmetraAcceptHeader)
```
HowardWolosky: The `symmetraAcceptHeader` is not relevant/acceptable for this set of API's.

HowardWolosky: Additionally, `Get-MediaAcceptHeader -MediaType Raw` would return back an AcceptHeader which included `application/vnd.github.v3.Raw+json` which is incorrect for this API.  You actually need `application/vnd.github.v3.Raw`

Shazwazza:
@HowardWolosky thanks for the notes! A bunch of this was copy/pasted (guilty!) so yup there's certainly some left over things lying around. I'll get back to this as soon as i can, cheers!

GitHubContents.ps1
```diff
@@ -0,0 +1,93 @@
...
+    Write-InvocationLog
+
+    $elements = Resolve-RepositoryElements -DisableValidation
+    $OwnerName = $elements.ownerName
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        'OwnerName' = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+        'Path' = $PSBoundParameters.ContainsKey('Path')
```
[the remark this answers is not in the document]
Shazwazza: For this logging it's not really possible to differentiate between whether it's a file or a directory since it's just a name and GitHub figures out if it's a file or directory so we'd only know that based on the results. What do you suggest in this case, should i just remove the Path parameter from being logged? Or do you have another idea?

Shazwazza:
Had pushed some changes and then started writing/running unit tests... unfortunately the `GitHubRepositoryForks.tests.ps1` test will delete any fork of this repo that you currently have :P maybe not the most ideal thing when trying to contribute, haha. 

I don't think there's another option here except for me to create a new fork and PR and close this one unless you know some other way?

Shazwazza:
Have pushed a new PR here https://github.com/microsoft/PowerShellForGitHub/pull/146

PR #146, opened 2020-02-04
Adding Get Repository.Contents functionality (take 2)

Shazwazza:
Original PR is here https://github.com/microsoft/PowerShellForGitHub/pull/143

But during testing my fork got deleted so this is a re-push of those files, plus tests and everything fixed up.

I still need to write a few more tests but please let me know if there's anything that is obvious that i need to fix up. 

Learning a lot about Powershell here :) got the static analysis running and tests 🎉


Shazwazza:
@HowardWolosky I've discovered something interesting regarding your [previous question](https://github.com/microsoft/PowerShellForGitHub/pull/143#pullrequestreview-341478845) of 

`Get-GitHubContents -OwnerName Microsoft -RepositoryName PowerShellForGitHub -Path README.md -MediaType Raw`

vs 

`Invoke-GHRestMethodMultipleResult -Description 'Getting Contents' -UriFragment /repos/Microsoft/PowerShellForGitHub/contents/README.md -AcceptHeader 'application/vnd.github.v3.Raw`

As it turns out there is actually 3 different results that GH will return.  The reason you are seeng a json structure returned with `application/vnd.github.v3.Raw` is because the casing of this header is invalid and it must be `application/vnd.github.v3.raw`. When `raw` is used, it returns the raw bytes of the file. When `html` is used it returns the converted md -> html bytes of the file. When anything else is used that is not an exact match, then the json document is returned which contains the base64 encoded version of the file. 

I'll update to support all 3.

HowardWolosky:
Thanks for the update.  Will review this PR later this week.

Shazwazza:
Sounds good, i've just pushed the latest update along with unit tests to cover the scenarios. There's a new switch `-ResultAsString` which will auto decode the resulting bytes or base64 string to a resulting string. There's tests to show this now.

GitHubContents.ps1
```diff
@@ -0,0 +1,122 @@
...
+        'OwnerName'      = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    $uriFragment = [String]::Empty
+    $description = [String]::Empty
+
+    $uriFragment = "/repos/$OwnerName/$RepositoryName/contents"
+
+    if ($PSBoundParameters.ContainsKey('Path')) {
```
HowardWolosky: formatting: Here and elsewhere -- this project puts braces on a newline by themselves.

GitHubContents.ps1
```diff
@@ -0,0 +1,122 @@
...
+        $Path = $Path.TrimStart("\", "/")
+        $uriFragment += "/$Path"
+        $description = "Getting content for $Path in $RepositoryName"
+    }
+    else {
+        $description = "Getting all content for in $RepositoryName"
+    }
+
+    $params = @{
+        'UriFragment'         = $uriFragment
```
HowardWolosky: formatting: Here and above (`$telemetryProperties`) - This project doesn't attempt to line up values this way.

GitHubCore.ps1
```diff
@@ -973,3 +973,36 @@ function Get-MediaAcceptHeader
 
     return ($acceptHeaders -join ',')
 }
+
+function Get-ContentMediaType
```
HowardWolosky: I completely agree that we should have a method that does this.  But even I'm confused about when we use `Get-ContentMediaType` vs `Get-MediaAcceptHeader`, _especially_ when all uses for the methods are to assign a value to `AcceptHeader`.  Is there any way that these can be unified?  I'm ok with the `ValidateSet` being a superset given that the caller is doing its own `ValidateSet` validation.  Maybe there is an additional `AsJson` switch that the existing callers would pass in for it to add the `+json` to the end of the header.  I'm not quite sure, and I'm open to suggestions.  It's just that it's too confusing to leave it as-is, especially for the next person who comes along needing to use one and being confused which they should use.

Tests/GitHubContents.tests.ps1
```diff
@@ -0,0 +1,151 @@
...
+try {
+    # Define Script-scoped, readonly, hidden variables.
+    @{
+        repoGuid       = [Guid]::NewGuid().Guid
+        readmeFileName = "README.md"
+    }.GetEnumerator() | ForEach-Object {
+        Set-Variable -Force -Scope Script -Option ReadOnly -Visibility Private -Name $_.Key -Value $_.Value
+    }
+
+    # Need this twice since we need to use the above var here
```
HowardWolosky: Need what twice?

Tests/GitHubContents.tests.ps1
```diff
@@ -0,0 +1,151 @@
...
+    @{
+        htmlOutput = "<div id=`"file`" class=`"md`" data-path=`"README.md`"><article class=`"markdown-body entry-content p-5`" itemprop=`"text`"><h1><a id=`"user-content-$repoGuid`" class=`"anchor`" aria-hidden=`"true`" href=`"#$repoGuid`"><svg class=`"octicon octicon-link`" viewBox=`"0 0 16 16`" version=`"1.1`" width=`"16`" height=`"16`" aria-hidden=`"true`"><path fill-rule=`"evenodd`" d=`"M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z`"></path></svg></a>$repoGuid</h1></article></div>"
```
HowardWolosky: I feel like this test is likely a little delicate because you're dependent on both the content of the default README.md that GitHub autogenerates to never change, as well as the way that it generates HTML from the MD file.  Ideally, we would make this test more robust by uploading our own file, and then downloading it and verifying that it round-tripped correctly.  Unfortunately, to do that would require the implementation of the [Create or update a file](https://developer.github.com/v3/repos/contents/#create-or-update-a-file) API which is beyond what you were intending to implement.

This test is solid as-is, but will likely need to be updated once that other API gets implemented in the module.

Tests/GitHubContents.tests.ps1
```diff
@@ -0,0 +1,151 @@
...
+#>
+
+# This is common test code setup logic for all Pester test files
+$moduleRootPath = Split-Path -Path $PSScriptRoot -Parent
+. (Join-Path -Path $moduleRootPath -ChildPath 'Tests\Common.ps1')
+
+try {
+    # Define Script-scoped, readonly, hidden variables.
+    @{
+        repoGuid       = [Guid]::NewGuid().Guid
```
HowardWolosky: Can you capitalize these (PascalCase) (and their uses) to indice that it's a constant vs a local var?

Tests/GitHubContents.tests.ps1
```diff
@@ -0,0 +1,151 @@
...
+
+            It "Should have the expected content" {
+                $readmeFileObject.contentAsString | Should be $rawOutput
+            }
+        }
+
+        Remove-GitHubRepository -Uri $repo.svn_url
+    }
+}
+finally {
```
HowardWolosky: formatting: Can you move this brace (and the one below in the `if`) to their own lines?  It's fine to leave the braces where they are for the `Context`/`Describe`/`It` expressions.

Tests/GitHubContents.tests.ps1
```diff
@@ -0,0 +1,151 @@
...
+<#
+.Synopsis
+   Tests for GitHubContents.ps1 module
+#>
+
+# This is common test code setup logic for all Pester test files
+$moduleRootPath = Split-Path -Path $PSScriptRoot -Parent
+. (Join-Path -Path $moduleRootPath -ChildPath 'Tests\Common.ps1')
+
+try {
```
HowardWolosky: formatting: Can you move this brace to its own line please?

GitHubCore.ps1
```diff
@@ -973,3 +973,36 @@ function Get-MediaAcceptHeader
...
+{
+<#
+    .DESCRIPTION
+        Returns a formatted AcceptHeader based on the requested MediaType for working with GitHub Content,
```
```suggestion
        Returns a formatted AcceptHeader based on the requested MediaType for working with GitHub Content.
```

GitHubContents.ps1
```diff
@@ -0,0 +1,122 @@
+function Get-GitHubContent {
```
HowardWolosky: Any reason you decided not to include the `Uri` ParameterSet here?  Is it because you thought that it might get too confused with the `Path` parameter?

GitHubContents.ps1
```diff
@@ -0,0 +1,122 @@
...
+
+    $telemetryProperties = @{
+        'OwnerName'      = (Get-PiiSafeString -PlainText $OwnerName)
+        'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    $uriFragment = [String]::Empty
+    $description = [String]::Empty
+
+    $uriFragment = "/repos/$OwnerName/$RepositoryName/contents"
```
HowardWolosky: There's no need to initialize `$uriFragment` to an empty string and then immediately assign this value.  Just assign this value in the initialization instead.

Shazwazza:
Hi @HowardWolosky thanks for the review, will fix up now. I think the auto-formatting in VS code is causing some of the formatting issues. I'll see what can be done about that and reply to your other questions inline.

Tests/GitHubContents.tests.ps1
Shazwazza: Sure thing, i was just following the convention from the `GitHubComments.tests.ps1`, in there these items are camelCase such as `defaultIssueTitle`. Or are you referring specifically to `repoGuid` since that is not a constant?

Tests/GitHubContents.tests.ps1
Shazwazza: The reason why there are 2x blocks defining "Script-scoped, readonly, hidden variables" is because the 2nd block requires a variable from the first block and it's not possible to use a variable within another variable declared in the same block. This syntax is new to me (sorry i'm not powershell wizard... yet ;) so I'm unsure what the best practice is here. Instead I can define the `repoGuid` as a normal variable and then use that within these script-scoped variables? Just let me know what the best practice is in this case. Thanks!

Tests/GitHubContents.tests.ps1
Shazwazza: Yes i totally agree! My plan was on the next iteration to build the create/update file API and then update these tests accordingly but wanted to get this functionality done first and then update these tests later once that functionality was implemented.

GitHubContents.ps1
Shazwazza: No reason, i just overlooked that functionality, i'll add it.

GitHubCore.ps1
Shazwazza: Yep i can modify the original method, i just didn't want to muck with the logic just in case but I totally agree that it's confusing to have both. I'll push an update.

Shazwazza:
Hi @HowardWolosky I've pushed a new round of updates:

* Streamlines Get-MediaAcceptHeader to support all scenarios and updates usages accordingly
* Adds .editorconfig and .vscode/settings.json to deal with auto formatting in vscode
* Adds Uri parameter and adds tests for it

HowardWolosky:
@Shazwazza - I completely missed that there was an update to this PR.  I thought I was still waiting on an update from you to review.  My apologies.  I will check out your updates this week.  So sorry about that.

Shazwazza:
Hi @HowardWolosky , any updates on this one? Thanks!

HowardWolosky:
@Shazwazza - Quite sorry about that. I promise to finish up the review this week.  Things got really hectic and harder to schedule time for recently with everything that's going on.  Hope you're staying safe, and I look forward to getting this change in shortly.  Thanks again for your patience.

Shazwazza:
No worries and hope you are well too. Hopefully I'll be able to get some more functionality added to this project in the coming months. 🤞

Tests/GitHubContents.tests.ps1
```diff
@@ -0,0 +1,176 @@
...
+{
+    # Define Script-scoped, readonly, hidden variables.
+    @{
+        repoGuid = [Guid]::NewGuid().Guid
+        readmeFileName = "README.md"
+    }.GetEnumerator() | ForEach-Object {
+        Set-Variable -Force -Scope Script -Option ReadOnly -Visibility Private -Name $_.Key -Value $_.Value
+    }
+
+    # Need this twice since we need to use the above var here
```
HowardWolosky:
```suggestion
    # Need two separate blocks to set constants because we need to reference a constant from the first block in this block.
```

Tests/GitHubContents.tests.ps1
HowardWolosky: ok, I've provided a an alternate way of writing that comment to make things a bit more clear.

Tests/GitHubContents.tests.ps1
```diff
@@ -0,0 +1,176 @@
...
+                # Replace newlines with empty for comparison
+                $readmeFileString.Replace("`n", "").Replace("`r", "") | Should be $htmlOutput
+            }
+        }
+
+        Context 'For getting html (string) file contents' {
+
+            $readmeFileString = Get-GitHubContent -OwnerName $script:ownerName -RepositoryName $repo.name -Path $readmeFileName -MediaType Html -ResultAsString
+
+            It "Should have the expected content" {
```
HowardWolosky: This is currently failing for me.

HowardWolosky:
```
DONE - Operation took 1 second(s) [Getting content for README.md in e9456bde-74de-4014-9543-1233fce5f6e1]
      [-] Should have the expected content 10ms
        Expected strings to be the same, but they were different.
        Expected length: 802
        Actual length:   798
        Strings differ at index 91.
        Expected: '...ntent p-5"...'
        But was:  '...ntent" ite...'
        115:                 $readmeFileString.Replace("`n", "").Replace("`r", "") | Should be $htmlOutput
        at <ScriptBlock>, F:\ACE\PowerShellForGitHub\Tests\GitHubContents.tests.ps1: line 115
```

Tests/GitHubContents.tests.ps1
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        Context 'For getting html (byte) file contents' {
+
+            $readmeFileBytes = Get-GitHubContent -OwnerName $script:ownerName -RepositoryName $repo.name -Path $readmeFileName -MediaType Html
+            $readmeFileString = [System.Text.Encoding]::UTF8.GetString($readmeFileBytes)
+
+            It "Should have the expected content" {
+                # Replace newlines with empty for comparison
+                $readmeFileString.Replace("`n", "").Replace("`r", "") | Should be $htmlOutput
```
HowardWolosky: This is currently failing for me:

```
DONE - Operation took 1 second(s) [Getting content for README.md in e9456bde-74de-4014-9543-1233fce5f6e1]
      [-] Should have the expected content 244ms
        Expected strings to be the same, but they were different.
        Expected length: 802
        Actual length:   798
        Strings differ at index 91.
        Expected: '...ntent p-5"...'
        But was:  '...ntent" ite...'
        105:                 $readmeFileString.Replace("`n", "").Replace("`r", "") | Should be $htmlOutput
        at <ScriptBlock>, F:\ACE\PowerShellForGitHub\Tests\GitHubContents.tests.ps1: line 105
```

Shazwazza:
@HowardWolosky Hi, i've pushed some changes. All tests are passing now. The problem is the hard coded file content was out of date. If we can get this part merged then I can work on create/update file apis so we don't have to worry about that. Let me know if all is good :)

HowardWolosky:
This has now been published out as part of [0.11.0](https://www.powershellgallery.com/packages/PowerShellForGitHub/0.11.0).  Thanks again for your contribution.

HowardWolosky:
@Shazwazza - You can see in #183 that these tests are failing again because the content has changed.  Are you still planning on working on some of the other Content API's so that we'd be able to set the content to make this less fragile?  You might be able to get some ideas from my Gist Draft PR #172.

PR #162, opened 2020-05-04
Adding functionality for Project Columns

jpomfret:
Covers: 
- List project columns
- Get a project column
- Create a project column
- Update a project column
- Delete a project column
- Move a project column

The tests don't run successfully without the Project code from PR #160, if you want I can rebase this branch with master once/if the core project code gets merged in?


GitHubProjectColumns.ps1
```diff
@@ -0,0 +1,381 @@
...
+function Get-GitHubProjectColumn
+{
+<#
+    .DESCRIPTION
+        Get the columns for a given Github Project.
+
+        The Git repo for this module can be found here: http://aka.ms/PowerShellForGitHub
+
+    .PARAMETER Project
+        Id of the project to retrieve a list of columns for.
```
HowardWolosky, line 13: nit: As a variable, it would be `Id`.  When mentioning it though, it's `ID` because it's an abbreviation.  Here and elsewhere.

GitHubProjectColumns.ps1
```diff
@@ -0,0 +1,381 @@
...
+
+        [string] $AccessToken,
+
+        [switch] $NoStatus
+    )
+
+    Write-InvocationLog
+
+    $telemetryProperties = @{}
+    $telemetryProperties['Name'] = Get-PiiSafeString -PlainText $Name
```
HowardWolosky, line 137: We don't get any value with having this.  Names aren't unique and so this doesn't provide any meaningful data.  We could capture the the `PiiSafeString` of Project, and then we could identify how many times a specific project created columns with this module, but knowing the name doesn't really get us much.

GitHubProjectColumns.ps1
```diff
@@ -0,0 +1,381 @@
...
+        If provided, this will be used as the AccessToken for authentication with the
+        REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
+
+    .PARAMETER NoStatus
+        If this switch is specified, long-running commands will run on the main thread
+        with no command line status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .EXAMPLE
```
HowardWolosky, line 252: Can you include an example that shows using `-Confirm:$false` [like was done here](https://github.com/microsoft/PowerShellForGitHub/blob/3fab72464e38cb573408add7e99d5a6bb0db2ea1/GitHubRepositories.ps1#L504)?

GitHubProjectColumns.ps1
```diff
@@ -0,0 +1,381 @@
...
+
+        Moves the project column with id 999999 to the first position.
+
+    .EXAMPLE
+        Move-GitHubProjectColumn -Column 999999 -Position Last
+
+        Moves the project column with id 999999 to the Last position.
+
+    .EXAMPLE
+        Move-GitHubProjectColumn -Column 999999 -Position After:888888
```
HowardWolosky, line 337: This syntax should be abstracted away from users.

I see two options here (maybe you see more):
* Have three mutually exclusive parameters: `[switch] $First`, `[switch] $Last`, `[int64] $After`, and add some error checking into the body to ensure that you either have `$true` for `$First` or `$Last` (since a user could pass in `-First:$false`), or a positive value for `$After`. Then build the appropriate body value from that.
* Have a `ValidateSet` with the options of First, Last, After, and then an additional parameter of `[int64] $After` and in-body validation that `$After` is only supplied `if ('after' -eq $Position)`.

I think the first option is cleaner.  I'm open to other suggestions.  I just _don't_ like users having to come up with the `After:888888` syntax themselves.

jpomfret: Totally agree. Thanks for the feedback, went ahead with option one.

jpomfret: Also - wondering if I should be clicking the resolve button when I fix it? Or should that be for you to resolve, once you review the adds?

Sorry if I'm doing this wrong!

GitHubProjectColumns.ps1
```diff
@@ -0,0 +1,407 @@
...
+
+        [string] $AccessToken,
+
+        [switch] $NoStatus
+    )
+
+    Write-InvocationLog
+
+    $telemetryProperties = @{}
+    $telemetryProperties['Name'] = Get-PiiSafeString -PlainText $Project
```
HowardWolosky, line 137:
```suggestion
    $telemetryProperties['Project'] = Get-PiiSafeString -PlainText $Project
```

jpomfret:
Awesome! Where do the tests run? Is that the 'Microsoft.PowerShellForGitHub' task that shows as expected?

HowardWolosky:
In theory, yes.  Those will fail right now however for two reasons:

1. The core changes aren't in master
2. There's an unresolved issue with the CI build on DevOps right now which I haven't figured out, so the tests always seem to fail for some reason.

I'll be running them manually for validation before merging, once the Core changes are in.

jpomfret:
Great - thanks for the explanation! I believe I have the core changes have been made and are ready for you.

HowardWolosky:
Looks like GitHub's web merge conflict resolver messed with the whitespace in the .psd1 file.  Can you take a look and get that fixed up? (Here and in the project cards PR as well)?

jpomfret:
> Looks like GitHub's web merge conflict resolver messed with the whitespace in the .psd1 file. Can you take a look and get that fixed up? (Here and in the project cards PR as well)?

I'm not sure what you mean by that, I pulled down the changes and the whitespace all looks ok... e.g.:

image

HowardWolosky:
You'll see the problem if you look at the ["Files Changed" tab](https://github.com/microsoft/PowerShellForGitHub/pull/162/files#diff-67359f329c0e4cf47e8d783fc344b069).

image

jpomfret:
I'm sorry - I'm still not following, if I look at the file it all looks lined up.

https://github.com/microsoft/PowerShellForGitHub/blob/68d297e51d9f114d161a44138ede46a125b706af/PowerShellForGitHub.psd1

I do see that it seems to think all the lines have changed, but I don't see why.

HowardWolosky:
> I do see that it seems to think all the lines have changed, but I don't see why.

Exactly.  That's what we want to avoid, because it destroys the ability to diff and clearly see what changed in the file.  Some sort of whitespace changed in it, but I can't tell what.  It doesn't appear to be the spaces themselves, nor the linefeeds.

The simplest way forward at this point I believe would be to copy the content that's currently in master, commit that (which would remove your changes), and then re-apply the changes you made and commit that again.  (Don't do it in one go because git may not notice the whitespace change and may perceive the file as not changed at all).

jpomfret:
Ahh!! Gotcha! 
Thanks and sorry, should be fixed here now.

HowardWolosky:
This is now available publicly in [0.12.0](https://www.powershellgallery.com/packages/PowerShellForGitHub/0.12.0).  Thanks so much for your contribution. 🎆

jpomfret:
Yes! This is great, thanks so much for all your coaching!

PR #167, opened 2020-05-12
Migrate REST API progress status to use Write-Progress

### LINKED ISSUE #119: Progress is Incorrectly Displayed
The helper function Wait-JobWithAnimation incorrectly handles progress, and this makes it look quite ugly indeed in the ISE (a spew of non-wrapped yellow lines).

The root of the problem is that you are creating your own Write-Progress.  Please use Write-Progress instead, as this is more standard and will be honored across more PowerShell hosts.

The issue reproduces everywhere, but is worst in the PowerShell ISE.
PowerShellForGitHub version: 0.8.0
$psVersionTable
PSVersion                      5.1.17763.503                                                                                  
PSEdition                      Desktop                                                                                        
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                        
BuildVersion                   10.0.17763.503                                                                                 
CLRVersion                     4.0.30319.42000                                                                                

----

To detail the fix:
Create an ID before the while loop (Line 57)

Change your Write-InteractiveHost to Write-Progress (line 94)
        Write-Progress "$($animationFrames[$($iteration % $($animationFrames.Length))])" "Elapsed: $([int]($iteration / $framesPerSecond)) second(s)" -PercentComplete $((($iteration % $($animationFrames.Length)) / $animationFrames.Length) * 100) -id $id

Complete the Progress message at the end (lines 101 & 108)

Get rid of your Write-InteractiveHost at the very end

HowardWolosky:
When this module was originally written, it used a special feature
of Write-Host in order to re-write the previously written line, in
order to show the user an indeterminate waiting indicator while the
module awaited a response from the server for a REST API request.

Unfortunately, this doesn't work well with all PowerShell hosts
(PowerShell ISE is one notable example).  This approach had been taken
originally as it had seemed odd to use Write-Progress (which displays
an absolute progress percentage) for something where we didn't absolutely
know the final duration.

This change migrates the module over to use Write-Progress, and simply
loops through the displayed percentage completed in the same manor that
it loops through and shows an animation.

This method should achieve the same desired goal (letting users know
that something is happening and that they just need to wait a bit
longer) in a more standards-approved way.

Resolve #119

### DIFF
--- a/Helpers.ps1
+++ b/Helpers.ps1
@@ -48,13 +48,11 @@ function Wait-JobWithAnimation
     $animationFrames = '|','/','-','\'
     $framesPerSecond = 9
 
-    # We'll wrap the description (if provided) in brackets for display purposes.
-    if ($Description -ne "")
-    {
-        $Description = "[$Description]"
-    }
-
+    $progressId = 1
     $iteration = 0
+    [int] $seconds = 0
+    $secondWord = 'seconds'
+
     while ($runningJobs.Count -gt 0)
     {
         # We'll run into issues if we try to modify the same collection we're iterating over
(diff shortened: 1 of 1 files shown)


StartAutomating:
Sorry for letting this fall of my radar.  

Taking a brief look at implementation, I don't see where you send  -Completed message to Write-Progress.

This will cause progress bars to stay visible while any other commands in the script execute.  E.g. if you ran a Get- that returned many results, and piped to an expensive Foreach-Object, your progress bar wouldn't go away until the Foreach-Object ended.

Sorry to let this drift off the radar again.  Thanks for improving your Write-Progress experience either way.

HowardWolosky:
Thanks for the feedback.  Putting out a fix now.

PR #176, opened 2020-05-21
Add tests to GitHubRepositories.tests.ps1

themilanfan:
Fixes #139 
Add tests for GitHubRepositories.ps1 functions

PR #192, opened 2020-05-28
GitHubRepositories: Add Additional Parameters

### LINKED ISSUE #189: New/Update-GitHubRepository: Add Support for delete_branch_on_merge API Property
### Description

The GitHub [Create](https://developer.github.com/v3/repos/#create-a-repository-for-the-authenticated-user)/[Update](https://developer.github.com/v3/repos/#update-a-repository) Repository APIs have a `delete_branch_on_merge` property which is not currently supported by the `New-GitHubRepository` and `Update-GitHubRepository` functions. This allows automatically deleting of head branches when pull requests are merged within the repo.

### Proposal

Add a `DeleteBranchOnMerge` switch parameter to both functions to add support for this feature.

### API Detail

```
delete_branch_on_merge | boolean | Either `true` to allow automatically deleting head branches when pull requests are merged, or `false` to prevent automatic deletion. Default: `false`.
```

X-Guardian:
#### Pull Request description

This PR adds the following parameters to the `New-GitHubRepository` and `Update-GitHubRepository` functions:

- DeleteBranchOnMerge
- IsTemplate

#### This Pull Request fixes the following issues

- Fixes #189


X-Guardian:
@HowardWolosky, can we trigger the CI on this PR?

HowardWolosky:
> can we trigger the CI on this PR?

Once #199 is resolved, I'll run CI on this.  Thanks for your patience (and many contributions).

HowardWolosky:
@X-Guardian -- don't you need the `application/vnd.github.nebula-preview+json` accept header for `visibility` and `application/vnd.github.baptiste-preview+json` for `template`?

X-Guardian:
Yes, @HowardWolosky , you are right. I've added these.

I haven't added any tests for these changes yet. I'll add these once #176 has been merged.

X-Guardian:
I don't think the pipeline triggered successfully.

X-Guardian:
I'm not seeing the test results on the Azure Pipelines job (there should be a 'test' tab). Also, any reason why the code coverage is configured to not be published when pulling from a fork?

HowardWolosky:
> I'm not seeing the test results on the Azure Pipelines job (there should be a 'test' tab).

That's been confusing me too.  Sometimes it shows up, sometimes it doesn't.  Here are [two](https://dev.azure.com/ms/PowerShellForGitHub/_build/results?buildId=84425&view=ms.vss-test-web.build-test-results-tab) [recent](https://dev.azure.com/ms/PowerShellForGitHub/_build/results?buildId=84441&view=ms.vss-test-web.build-test-results-tab) runs where it _did_ show up.

> Also, any reason why the code coverage is configured to not be published when pulling from a fork?

I really can't think of any.  I'll change the yaml template shortly.

X-Guardian:
> > I'm not seeing the test results on the Azure Pipelines job (there should be a 'test' tab).
> 
> That's been confusing me too. Sometimes it shows up, sometimes it doesn't. Here are [two](https://dev.azure.com/ms/PowerShellForGitHub/_build/results?buildId=84425&view=ms.vss-test-web.build-test-results-tab) [recent](https://dev.azure.com/ms/PowerShellForGitHub/_build/results?buildId=84441&view=ms.vss-test-web.build-test-results-tab) runs where it _did_ show up.
> 

I don't see the test tab on either of those runs.

I'm seeing the same issue with other open-source projects, so I've been digging around and it looks like Microsoft are now hiding the test tab for non-organization members. See: https://developercommunity.visualstudio.com/content/problem/1045997/tests-tab-not-visible-for-logged-out-users-on-open.html. I hope they change there minds on this!

GitHubRepositories.ps1
```diff
@@ -106,6 +115,9 @@ function New-GitHubRepository
 
         [switch] $Private,
 
+        [ValidateSet('Public', 'Private', 'Internal')]
+        [string] $Visibility,
```
HowardWolosky, line 119: In an ideal world, we'd only have this and wouldn't have `[switch] $Private`, because this can lead to confusion, but dropping `[switch] $Private` would be a breaking change.  So, let's just do some safe-guarding.

1. Please add a check in the body that's something similar to:
```powershell
    if ($PSBoundParameters.ContainsKey('Private'))
    {
        Write-Log -Level Warning -Message 'The Private switch is being replaced by the Visibility parameter.  Please update any scripts/usage that you may have to migrate to the Visibility parameter, as the Private switch will be replaced in a future update.'

        if ($PSBoundParameters.ContainsKey('Visibility'))
        {
            if (($Private -and ($Visibility -ne 'Private')) -or ((-not $Private) -and ($Visibility -ne 'Public')))
            {
                $message = 'The Private and Visibility parameters control the same repo property.  If both are specified, their values must be the same.'
                Write-Log -Message $message -Level Error
                throw $message
            }
        }
    }
```

2. Please enhance the `Visibility` CBH description to explain that `Internal` is only available `if your organization is associated with an enterprise account using GitHub Enterprise Cloud or GitHub Enterprise Server 2.20+`, and that it controls the same property as the `Private` switch.

3. Please update the `Private` switch CBH description to explain that it controls the same property as `Visibility` and that only one should be used.

GitHubRepositories.ps1
```diff
@@ -156,19 +172,23 @@ function New-GitHubRepository
     if ($PSBoundParameters.ContainsKey('LicenseTemplate')) { $hashBody['license_template'] = $LicenseTemplate }
     if ($PSBoundParameters.ContainsKey('TeamId')) { $hashBody['team_id'] = $TeamId }
     if ($PSBoundParameters.ContainsKey('Private')) { $hashBody['private'] = $Private.ToBool() }
+    if ($PSBoundParameters.ContainsKey('Visibility')) { $hashBody['visibility'] = $Visibility }
```
HowardWolosky, line 175: I believe that you'll have to keep this lower:

```suggestion
    if ($PSBoundParameters.ContainsKey('Visibility')) { $hashBody['visibility'] = $Visibility.ToLower() }
```

GitHubRepositories.ps1
HowardWolosky: Similar feedback for the `Update-GitHubRepository`

X-Guardian: Done

GitHubRepositories.ps1
X-Guardian: Done, and I've changed the same in `Update-GitHubRepository`.

GitHubRepositories.ps1
```diff
@@ -131,6 +151,24 @@ function New-GitHubRepository
         'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
     }
 
+    if ($PSBoundParameters.ContainsKey('Private'))
+    {
+        Write-Log -Level Warning -Message ('The Private switch is being replaced by the Visibility parameter.  ' +
+            'Please update any scripts/usage that you may have to migrate to the Visibility parameter, ' +
+            'as the Private switch will be replaced in a future update.')
```
HowardWolosky, line 158: Your usage of deprecated above is better for this message.

```suggestion
            'as the Private switch will be deprecated in a future update.')
```

GitHubRepositories.ps1
```diff
@@ -817,6 +880,24 @@ function Update-GitHubRepository
...
```
HowardWolosky, line 887:
```suggestion
            'as the Private switch will be deprecated in a future update.')
```

GitHubRepositories.ps1
```diff
@@ -131,6 +151,24 @@ function New-GitHubRepository
         'RepositoryName' = (Get-PiiSafeString -PlainText $RepositoryName)
     }
 
+    if ($PSBoundParameters.ContainsKey('Private'))
+    {
+        Write-Log -Level Warning -Message ('The Private switch is being replaced by the Visibility parameter.  ' +
```
HowardWolosky, line 156: This message also got me thinking..._we_ have scripts/usage that needs to migrate as well.  Can you also update the `.EXAMPLES` in the CBH within this script as well as the few usages of `-Private` in the tests (I found two instances)?

X-Guardian: Changed examples and one reference in `GitHubRepositories.tests.ps1`. Couldn't see a second reference in the tests?

GitHubRepositories.ps1
HowardWolosky: Looks like you missed the one in GitHubProjects.tests.ps1:

https://github.com/microsoft/PowerShellForGitHub/blob/00f099fa81d25908735dc32d86736180f6209d34/Tests/GitHubProjects.tests.ps1#L204

GitHubRepositories.ps1
```diff
@@ -760,8 +816,12 @@ function Update-GitHubRepository
     .EXAMPLE
         Update-GitHubRepository -OwnerName Microsoft -RepositoryName PowerShellForGitHub -Description 'The best way to automate your GitHub interactions'
 
+        Changes the description of the specified repository.
+
     .EXAMPLE
-        Update-GitHubRepository -Uri https://github.com/PowerShell/PowerShellForGitHub -Private:$false
+        Update-GitHubRepository -Uri https://github.com/PowerShell/PowerShellForGitHub -Visibility 'Private'
+
+        Changes the visibility of the specified repository to 'Private'.
```
HowardWolosky, line 824: I think this one is a bad example that won't work.  The example that you're replacing was changing a repo to to be _Public_ (`-Private:$false`).  I don't think that GitHub _lets_ you change a repo to be Private once it's already public.

```suggestion
        Update-GitHubRepository -Uri https://github.com/PowerShell/PowerShellForGitHub -Visibility 'Public'

        Changes the visibility of the specified repository to be public.
```

X-Guardian: You can do it from the GUI, though I've just tried it from the function, and it didn't error, but hasn't changed the repo's state.

By the way, I am receiving the following error when calling any functions from the current master branch:

```
Microsoft.PowerShell.Commands.HttpResponseException: Response status code does not indicate success: 400 (Invalid instrumentation key).
   at System.Management.Automation.MshCommandRuntime.ThrowTerminatingError(ErrorRecord errorRecord)
```

GitHubRepositories.ps1
X-Guardian: That's a different function though...

HowardWolosky: <sigh> Completely missed that.  Apologies.

GitHubRepositories.ps1
HowardWolosky: Yeah, the other reference was `Set-GitHubProject` ... I didn't look at the code carefully enough.

GitHubRepositories.ps1
HowardWolosky: Interesting. I could have sworn that once it was public, you couldn't make it private again...maybe that's only if there are forks.  I still think it would be better to keep the example going to public than to private.

X-Guardian: In the GUI it comes up with a big warning with a list of features that will be lost going from Public to Private. I changed the example anyway.

HowardWolosky: As for the error, I can't reproduce it.  It _sounds_ like it's related to #186, but it's not repro'ing for me.  My best guess is that you've modified the `ApplicationInsightsKey` configuration property, and so you're seeing that error when it's trying to report the telemetry.

To verify:
```powershell
Get-GitHubConfiguration -Name ApplicationInsightsKey
```

The expected value should be:
https://github.com/microsoft/PowerShellForGitHub/blob/d1bd976d70cc975dfd247f9ad2bace58a465c7da/GitHubConfiguration.ps1#L611

If you're not seeing that, then you have a bunch of options ahead of you to fix it:
1. `Set-GitHubConfiguration -ApplicationInsightsKey '<that value>'`  (this will make it work as expected)
2. `Reset-GitHubConfiguration` (this will lose all your current settings and get all the default ones)
3. `Set-GitHubConfiguration -DisableTelemetry` (this will ignore that invalid ApplicationInsightsKey because it will just stop reporting telemetry)

X-Guardian: OK `Reset-GitHubConfiguration` fixed it, and I know how I broke it. I ran `Set-GitHubConfiguration` and missed out the 'dash' prefix on the parameter name I wanted to set, which because none of the functions have `PositionalBinding=$false` set, overwrote the first two properties (`ApiHostName` and `ApplicationInsightKey`).
My opinion is functions should always have `PositionalBinding=$false` set and if required, the `position` property set on individual parameters.

HowardWolosky: Didn't even know `PositionalBinding=$false` was a thing.  You're not the first one to get bitten on a similar type of issue for `Set-GitHubConfiguration`.  Of all the functions, that one is ripe for that value.  My only concern is that, [based on the docs](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.cmdletbindingattribute.positionalbinding), that's not available in PS 4.0 which we still need to support.

X-Guardian: I thought `PositionalBinding` was implemented with Advanced Functions back in PowerShell 3, but I couldn't say for sure. Why does the module still need to support PowerShell 4?

HowardWolosky: Some internal systems that this module is used on are still running on older versions of Windows Server which are still using PS4, and it's not within our power to update those servers nor to install PS7Core.  At some point those servers will get updated, but until then:

https://github.com/microsoft/PowerShellForGitHub/blob/d1bd976d70cc975dfd247f9ad2bace58a465c7da/PowerShellForGitHub.psd1#L48-L49

HowardWolosky: I've opened #219 to track this.

HowardWolosky:
[This CI failed](https://dev.azure.com/ms/PowerShellForGitHub/_build/results?buildId=85018&view=ms.vss-test-web.build-test-results-tab&runId=2400254&resultId=100178&paneView=debug) due to a single related test:

```
Expected collection $null to contain '40c9069e-02ae-4234-a1a2-fe5f21308e2e', but it was not found.
```

```
at <ScriptBlock>, D:\a\1\s\Tests\GitHubRepositories.tests.ps1: line 37
37:                 $privateRepo.Name | Should BeIn $privateRepos.Name
```

I can repro that failure locally as well, but only with your change.

HowardWolosky:
I also confirmed by manually using it that, despite the body looking correct, the request is still creating a _public_ repo:

```powershell
Set-GitHubConfiguration -LogRequestBody
New-GitHubRepository -RepositoryName "testp" -AutoInit -Visibility "Private"
```

log
```
2020-06-02 15:16:20 : Howard : VERBOSE : Accessing [Post] https://api.github.com/user/repos [Timeout = 0)]
2020-06-02 15:16:21 : Howard : VERBOSE : Request includes a body.
2020-06-02 15:16:21 : Howard : VERBOSE : {
  "visibility": "private",
  "name": "testp",
  "auto_init": true
}
```

result...

```powershell
(Get-GitHubRepository -OwnerName PowerShellForGitHubTeam -RepositoryName testp).Visibility
public
```

X-Guardian:
I think it is a problem with the API. If you specify the `OrganizationName` parameter as well as `Visibility`, the Repo is successfully created as private.

How about we remove the `Visibility` parameter from this PR and implement it later? I was particularly interested in having the `DeleteBranchOnMerge` parameter.

HowardWolosky:
> How about we remove the `Visibility` parameter from this PR and implement it later? I was particularly interested in having the `DeleteBranchOnMerge` parameter.

That works for me.  Would you mind creating a tracking issue for `Visibility` though, along with a reference to this PR and what we've learned so far... Bonus points if you want to report the issue to GitHub as well so that they can either fix the API (or fix the documentation to indicate that it's organization-only).

HowardWolosky:
>OK, I've removed the `Visibility` parameter. I'll raise the issue here: >https://github.community/c/github-api-development-and-support.
>
>I'll raise another PR with the `Visiblity` parameter code, so we don't lose it.
>
>Reminder: this PR still needs additional tests to be added, but I'm waiting for #176 to be merged.

Thanks for all that.  #176 should be getting merged in today.

X-Guardian:
I've updated the GitHubRepositories tests with the following:

- Added global CodeAnalysis suppression of `PSUseDeclaredVarsMoreThanAssignments` and removed redundant work-around code as this check serves little purpose with Pester.
- Coverage of all `New-GitHubRepository` lines apart from `$TeamId` processing, which needs a 'New-GitHubTeam` function.
- Coverage of all `Update-GitHubRepository` lines apart from `$DefaultBranch` processing, which needs a `New-GitHubRepositoryBranch` function. (to be added with PR #200).

X-Guardian:
By the way, you have added a `breaking change` label to this PR, but I don't think it is?

HowardWolosky:
> By the way, you have added a `breaking change` label to this PR, but I don't think it is?

Removed.  Thanks for the reminder.  That was there back when we were prepping to deprecate the `Private` switch.

X-Guardian:

X-Guardian:
Hi @HowardWolosky, are we able to get this PR reviewed and merged?

X-Guardian:
I would suggest adding `Set-StrictMode -Version 1.0` to the tests to pick up on mistyped/incorrectly scoped variable names at runtime. Much better than `PSUseDeclaredVarsMoreThanAssignments` which is fundamentally flawed.

HowardWolosky:
> I would suggest adding `Set-StrictMode -Version 1.0` to the tests to pick up on mistyped/incorrectly scoped variable names at runtime. Much better than `PSUseDeclaredVarsMoreThanAssignments` which is fundamentally flawed.

Brilliant!  It won't identify variables that are assigned to but are never used, but it _will_ at least capture when a variable is accessed that wasn't previously created.  I do think that we're losing _a little_ something by not identifying variables that we are saving but not using (which might point to a miswritten test), but this will make the tests more manageable (and readable) in general by avoiding the workaround I was previously using.

I tested it out by putting it in Tests/Common.ps1, and then I inserted the following to the top of a test file:

```powershell
$foo = $bar
```
and got the following error immediately upon executing the test:
```
[-] Error occurred in test script '.\tests\GitHubUsers.tests.ps1' 0ms
  RuntimeException: The variable '$bar' cannot be retrieved because it has not been set.
  at <ScriptBlock>, C:\git\PowerShellForGitHub\tests\GitHubUsers.tests.ps1: line 14
  at <ScriptBlock>, C:\Program Files\WindowsPowerShell\Modules\Pester\4.10.1\Pester.psm1: line 1111
  at Invoke-Pester<End>, C:\Program Files\WindowsPowerShell\Modules\Pester\4.10.1\Pester.psm1: line 1137 
```

Will be making use of this in the updated pipeline tests.

Thanks!

PR #207, opened 2020-05-31
Speed up Pester tests by using -NoStatus

### LINKED ISSUE #206: Suggestion: Add 'Set-GitHubConfiguration -DefaultNoStatus' to All Pester Tests
Currently the full suite of Pester tests take between 20 to 25 minutes to run per environment within the CI pipeline. There are currently three environments being tested (Windows, Linux and MacOS) meaning the full CI takes about 1 hour 15 mins to run.

Locally running the `GitHubRepositories.tests` Pester tests and adding `Set-GitHubConfiguration -DefaultNoStatus` in a `BeforeAll` block for each `Describe` block, reduces the time for the `GitHubRepositories` tests to run from approx 60 secs to 25 secs.

Extrapolating this up, applying this change to all tests may reduce the full test suite run from 1 hour 15 mins to approx 30 mins.

I've also noticed that the Azure DevOps Pipeline raw log for the 'Run Unit Tests via Pester` step are 'corrupted' by the status output on [Linux](https://dev.azure.com/ms/341bcd7d-fac3-430a-a077-9c95d49a8ba5/_apis/build/builds/84257/logs/23) and [MacOS](https://dev.azure.com/ms/341bcd7d-fac3-430a-a077-9c95d49a8ba5/_apis/build/builds/84257/logs/35), unlike on [Windows](https://dev.azure.com/ms/341bcd7d-fac3-430a-a077-9c95d49a8ba5/_apis/build/builds/84257/logs/10). Making this change would also resolve this.

HowardWolosky:
Suppressing Status (e.g. using `-NoStatus`) reduces the overall runtime of the UT's across all 3 platforms (Windows, Linux, Mac) by 25-30 min.  It also has the side-effect of improving the readability of the raw logs on CI for Linux and Mac.

[Most recent run](https://dev.azure.com/ms/PowerShellForGitHub/_build/results?buildId=84315) without this change: 1h 19m
[Run with this change](https://dev.azure.com/ms/PowerShellForGitHub/_build/results?buildId=84357) without this change: 54m

Resolves #206

### DIFF
--- a/Tests/Common.ps1
+++ b/Tests/Common.ps1
@@ -89,6 +89,7 @@ function Initialize-CommonTestSetup
     Reset-GitHubConfiguration
     Set-GitHubConfiguration -DisableTelemetry # We don't want UT's to impact telemetry
     Set-GitHubConfiguration -LogRequestBody # Make it easier to debug UT failures
+    Set-GitHubConfiguration -DefaultNoStatus # Status corrupts the raw CI logs for Linux and Mac, and makes runs take slightly longer.
 }
 
 Initialize-CommonTestSetup


X-Guardian:
OK, my extrapolation was a bit optimistic, but a 25 minute reduction is not bad!

HowardWolosky:
> OK, my extrapolation was a bit optimistic, but a 25 minute reduction is not bad!

Absolutely!  This is still a great improvement, all with a single line of code.  So, thanks so much for the suggestion!

Issue #219, opened 2020-06-02
Consider adding PositionalBinding=$false for some functions


HowardWolosky:
#### Feature Idea Summary
This comes out of a [comment thread](https://github.com/microsoft/PowerShellForGitHub/pull/192#discussion_r434160733) in #192.  We've seen a number of times that users who forget to provide a `-` for the parameter name when calling `Set-GitHubConfiguration` end up mistakingly changing the configuration for one or more incorrect settings.  The idea here is to use [PositionalBinding=$false](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.cmdletbindingattribute.positionalbinding) to disable positional binding for at least that function, if not others.

#### Feature Idea Additional Details
The concern with this idea is that the docs seem to imply that this started with PS5, but this module needs to continue to support PS4+ due to some internal requirements.  If someone can verify that PS4 supports this configuration property, then we'd be safe to continue with it.  That verification _must_ happen first though.

#### Requested Assignment
I'm just suggesting this idea, but don't want to implement it.

#### Operating System
```
OsName               : Microsoft Windows 10 Pro
OsOperatingSystemSKU : 48
OsArchitecture       : 64-bit
WindowsVersion       : 1909
WindowsBuildLabEx    : 18362.1.amd64fre.19h1_release.190318-1202
OsLanguage           : en-US
OsMuiLanguages       : {en-US}
```
#### PowerShell Version
```
Name                           Value
----                           -----
PSVersion                      7.0.1
PSEdition                      Core
GitCommitId                    7.0.1
OS                             Microsoft Windows 10.0.18363
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```
#### Module Version
```
Running: 0.14.0
Installed:
```

X-Guardian:
I setup this test script on a Windows 2012 R2 server running PowerShell 4.0:

```powershell
[CmdletBinding(PositionalBinding=$false)]
param(
    [Parameter(Position=1)]
    [string] $Parm1,

    [Parameter()]
    [string] $Parm2,

    [Parameter()]
    [string] $Parm3
)
```

Here is the output of `Get-Help` for the script, showing the positional binding working:

```powershell
[...]

PR #222, opened 2020-06-06
GitHubTeams: Add New/Update/Remove Functions

X-Guardian:
#### Description

This PR adds the following functions to the `GitHubTeams` module:

- `Add-GitHubTeam`
- `Update-GitHubTeam`
- `Remove-GitHubTeam`

It also adds the `TeamName` parameter to the `Get-GitHubTeam` function to allow getting team details by team name, which is used by the new functions.

100% coverage Pester tests have also been added for the following functions:

- `Get-GitHubTeam`
- `Add-GitHubTeam`
- `Update-GitHubTeam`
- `Remove-GitHubTeam`

Positional Binding has been set as `false` for the three functions, and `Position` attributes added to the function's mandatory parameters.

#### Issues Fixed

None

#### References

[GitHub Teams API](https://developer.github.com/v3/teams/)

#### Checklist

- [x] [Static analysis](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#static-analysis)
is reporting back clean.


X-Guardian:

X-Guardian:

X-Guardian:
I've added the `-Force` switch to the `Remove-GitHubTeams` function as detailed in issue #218.

X-Guardian:
@HowardWolosky , this PR is ready for review.

HowardWolosky:
Thanks @X-Guardian ... holding off on this one for another few days until the big pipelining change is in so as not to accumulate further debt.  I'll start a review on this soon, but there will be just a little bit of additional work needed to align the new functions to the way pipelining will be working with the rest of the module before it gets merged in.

As an aside, I'm in agreement on removing implicit positional parameters on these and future functions per our discussion around #219.

GitHubTeams.ps1
```diff
@@ -150,11 +158,41 @@ filter Get-GitHubTeam
...
+        $team = $result | Where-Object -Property name -eq $TeamName
+
+        if ($null -eq $team)
+        {
+            throw "Team '$TeamName' not found"
+        }
+        else
+        {
+            $uriFragment = "/teams/$($team.Id)"
+            $description = "Getting team $($team.Id)"
```
HowardWolosky, line 178: Technically it's a lower-case property:

```suggestion
            $uriFragment = "/teams/$($team.id)"
            $description = "Getting team $($team.id)"
```


The added one is SentenceCase: `$team.TeamId`

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
 }
 
+function New-GitHubTeam
+{
+<#
+    .SYNOPSIS
+        Creates a team within an organization on GitHub.
+
+    .DESCRIPTION
+        Creates a team or within an organization on GitHub.
```
HowardWolosky, line 324:
```suggestion
        Creates a team within an organization on GitHub.
```

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+
+    .OUTPUTS
+        PSCustomObject
+
+    .EXAMPLE
+        New-GitHubTeam -OrganizationName PowerShell -TeamName 'Developers'
+
+        Creates a new GitHub team called 'Developers' in the 'PowerShell' organization.
+#>
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSShouldProcess', '',
```
HowardWolosky, line 367: You don't have to change these, but the `SuppressMessageAttribute`s are one of the few exceptions in my book for exceeding the line length requirements.

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+    .PARAMETER OrganizationName
+        The name of the organization to create the team in.
+
+    .PARAMETER TeamName
+        The name of the team.
+
+    .PARAMETER Description
+        The description for the team.
+
+    .PARAMETER Maintainers
```
HowardWolosky, line 337: We tend to follow the PowerShell guidance of using the singular noun or verb for names (see any of the label or issue methods, for example).

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+        [ValidateNotNullOrEmpty()]
+        [string] $OrganizationName,
+
+        [Parameter(Mandatory, Position = 2)]
+        [ValidateNotNullOrEmpty()]
+        [string] $TeamName,
+
+        [string] $Description,
+
+        [string[]] $Maintainers,
```
HowardWolosky, line 389: If this is really supposed to be the ID, then this should be `[int64[]]` as opposed to `[string[]]`.

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+
+        [string[]] $RepositoryName,
+
+        [ValidateSet('Secret','Closed')]
```
HowardWolosky, line 393: nit:
```suggestion
        [ValidateSet('Secret', 'Closed')]
```

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+            Whatif = $false
+            Confirm = $false
+        }
+        if ($PSBoundParameters.ContainsKey('AccessToken'))
+        {
+            $getGitHubTeamParms['AccessToken'] = $AccessToken
+        }
+        if ($PSBoundParameters.ContainsKey('NoStatus'))
+        {
+            $getGitHubTeamParms['NoStatus'] = $NoStatus
```
HowardWolosky, line 434: This won't work right (I know because I made the same mistake in my Releases PR).  You'll see the bug during Pester test runs when `DefaultNoStatus` is set to `$true`.  You need to use `Resolve-ParameterWithDefaultConfigurationValue -Name NoStatus -ConfigValueName DefaultNoStatus`

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+        }
+        $team = Get-GitHubTeam @getGitHubTeamParms
```
HowardWolosky, line 436: nit: please add a line feed here after the close of the expression

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+        Method = 'Post'
+        Description =  "Creating $TeamName"
+        AccessToken = $AccessToken
+        TelemetryEventName = $MyInvocation.MyCommand.Name
+        TelemetryProperties = $telemetryProperties
+        NoStatus = (Resolve-ParameterWithDefaultConfigurationValue `
+            -Name NoStatus -ConfigValueName DefaultNoStatus)
+    }
+
+    return Invoke-GHRestMethod @params
```
HowardWolosky, line 453: you'll need to pipe to `Add-GitHubTeamAdditionalProperties` here.

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+}
+
+function Update-GitHubTeam
```
HowardWolosky, line 456: please update to be a `filter` and all other relevant pipeline-supporting changes.  This would definitely benefit from filter input, being able to pass-in a Team object...which means that the slug needs to be an optional parameter set.

HowardWolosky, line 456: Relevant to #228, please change to `Set` verb.

```suggestion
function Set-GitHubTeam
```

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+    }
+    if ($PSBoundParameters.ContainsKey('AccessToken'))
+    {
+        $getGitHubTeamParms['AccessToken'] = $AccessToken
+    }
+    if ($PSBoundParameters.ContainsKey('NoStatus'))
+    {
+        $getGitHubTeamParms['NoStatus'] = $NoStatus
+    }
+    $team = Get-GitHubTeam @getGitHubTeamParms
```
HowardWolosky, line 553: nit: please add a linefeed here

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+        Whatif = $false
+        Confirm = $false
+    }
+    if ($PSBoundParameters.ContainsKey('AccessToken'))
+    {
+        $getGitHubTeamParms['AccessToken'] = $AccessToken
+    }
+    if ($PSBoundParameters.ContainsKey('NoStatus'))
+    {
+        $getGitHubTeamParms['NoStatus'] = $NoStatus
```
HowardWolosky, line 551: As before, you need to use `Resolve-ParameterWithDefaultConfigurationValue` here and elsewhere, otherwise this won't honor that config preference,

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+    .PARAMETER Maintainers
+        A list of GitHub IDs for organization members who will become team maintainers.
+
+    .PARAMETER RepositoryName
+        A list of repositories to add the team to.
+
+    .PARAMETER Privacy
+        The level of privacy this team should have.
+
+    .PARAMETER ParentTeamName
```
HowardWolosky, line 346: I think it would make sense for the parent team to be targetable by the pipeline once implemented.  So, specifying either the ParentTeamName (not via pipeline) or the ParentTeamSlug (via `PipelineByPropertyName`)

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+        TelemetryEventName = $MyInvocation.MyCommand.Name
+        TelemetryProperties = $telemetryProperties
+        NoStatus = (Resolve-ParameterWithDefaultConfigurationValue `
+            -Name NoStatus -ConfigValueName DefaultNoStatus)
+    }
+
+    return Invoke-GHRestMethod @params
+}
+
+function Remove-GitHubTeam
```
HowardWolosky, line 599: as before, please migrate to `filter` with pipeline support, including having the slug be passable from the pipleline by property name.

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+{
+<#
+    .SYNOPSIS
+        Removes a team within an organization on GitHub.
```
HowardWolosky, line 603:
```suggestion
        Removes a team from an organization on GitHub.
```

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+
+    .DESCRIPTION
+        Removes a team within an organization on GitHub.
```
HowardWolosky, line 606:
```suggestion
        Removes a team from an organization on GitHub.
```

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', '',
+        Justification = 'One or more parameters (like NoStatus) are only referenced by helper
+        methods which get access to it from the stack via Get-Variable -Scope 1.')]
+    [CmdletBinding(
+        SupportsShouldProcess,
+        PositionalBinding = $false
+    )]
+    param
+    (
+        [Parameter(Mandatory, Position = 1)]
```
HowardWolosky, line 379: Here and elsewhere, since the pipeline change, the module is now consistent with having the `Parameter` attribute separate its properties on individual lines if there are more than one.

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+    }
+    if ($PSBoundParameters.ContainsKey('AccessToken'))
+    {
+        $getGitHubTeamParms['AccessToken'] = $AccessToken
+    }
+    if ($PSBoundParameters.ContainsKey('NoStatus'))
+    {
+        $getGitHubTeamParms['NoStatus'] = $NoStatus
+    }
+    $team = Get-GitHubTeam @getGitHubTeamParms
```
HowardWolosky, line 690: nit: please add a line feed

Tests/GitHubTeams.tests.ps1
```diff
@@ -11,13 +11,516 @@
...
+
+        Context 'When getting a GitHub Team by organization' {
+            BeforeAll {
+                $repoName = [Guid]::NewGuid().Guid
+
+                $newGithubRepositoryParms = @{
+                    RepositoryName = $repoName
+                    OrganizationName = $script:organizationName
+                }
+                $repo = New-GitHubRepository @newGitHubRepositoryParms
```
HowardWolosky, line 41: nit: line feed, here and below, separating the params from the execution line.  I generally prefer a linefeed after any closing expression, but with some exceptions have been allowing that to be avoided in the case of repeated expressions that are configuring a params hashtable.

GitHubTeams.ps1
```diff
@@ -276,6 +314,407 @@ filter Get-GitHubTeamMember
...
+    .PARAMETER TeamName
+        The name of the team.
+
+    .PARAMETER Description
+        The description for the team.
+
+    .PARAMETER Maintainers
+        A list of GitHub IDs for organization members who will become team maintainers.
+
+    .PARAMETER RepositoryName
```
HowardWolosky, line 340: We don't want users to be "surprised" by unexpected differences in behavior.  Throughout the rest of the module `RepositoryName` refers specifically to just the individual name of the repo, _not_ its full name which is what this parameter is _actually_ expecting.  I would suggest renaming this to something like `RepositoryFullName` and then updating the description of this comment to be a lot more clear on what is being expected.

Given this new property needed for API interaction, you may also want to update `Add-GitHubRepositoryAdditionalProperties` to also set a `RepositoryFullName` property that this could be then retrieve via the pipeline by property name.  If you do so, leave this as-is as a function, but then change to using the `begin/process/end` format to collect all of the repository full names so that you can use them in a single execution of the API.

HowardWolosky: In fact, you probably want to do that anyway so that you could also support the pipelining in of multiple user objects if the user is trying to specify the maintainers via the pipeline instead.

Tests/GitHubTeams.tests.ps1
```diff
@@ -11,13 +11,516 @@
...
+
+                $getGitHubTeamParms = @{
+                    OrganizationName = $script:organizationName
+                    TeamName = $teamName
+                }
+                $team = Get-GitHubTeam @getGitHubTeamParms
+            }
+
+            It 'Should return an object of the correct type' {
+                $team | Should -BeOfType PSCustomObject
```
HowardWolosky, line 66: I'll just mention once here that you'll want to update the type check to be more like `$team.PSObject.TypeNames[0] | Should -Be 'GitHub.Team'` and you may want to check for additional properties as well.  Here and elsewhere.

Tests/GitHubTeams.tests.ps1
```diff
@@ -11,13 +11,516 @@
...
+                $team.description | Should -Be $description
+                $team.organization.login | Should -Be $script:organizationName
+                $team.parent | Should -BeNullOrEmpty
+                $team.members_count | Should -Be 1
+                $team.repos_count | Should -Be 1
+                $team.privacy | Should -Be $privacy
+            }
+
+            AfterAll {
+                if ($team)
```
HowardWolosky, line 80: It's not clear to me that this conditional is needed.  I believe that AfterAll is going to run even if the internal tests fail.  And if BeforeAll fails, either AfterAll will also fail and error, or it won't run at all.  Either way, I'm not sure it adds much additional value, plus this will likely be able to be simplified to just `$team | Remove-GitHubTeam -Force` once pipelining is done.  Simple, easy to read one-liner.

Here and elsewhere (except for the Remove-GitHubTeam tests themselves of course)

GitHubTeams.ps1
HowardWolosky: My name is `HowardWolosky`.  My id is `22158599`.  The [documentation](https://developer.github.com/v3/teams/#create-a-team) is confusing by incorrectly referring to strings (names) as ID's.

GitHubTeams.ps1
HowardWolosky: You're totally right.  My mistake.  With my issue, I was [directly referring to `$NoStatus`](https://github.com/microsoft/PowerShellForGitHub/pull/177/commits/03ec7d969a7b17b5a3de1373f3117627124df464#diff-b2345e64c8f3c38d6742b77af79129a0L951) which caused the issue.  You're guarding yourself here with the `ContainsKey` check.  So, you won't run afoul of the same issue.  You _could_ choose to use `Resolve-ParameterWithDefaultConfigurationValue` to avoid the check still, but keeping the check there still maintains the same end result.

GitHubTeams.ps1
HowardWolosky: Your right there.  At the time that I wrote that I must have forgotten that `TeamName` / 'TeamSlug' was already a parameter.

GitHubTeams.ps1
HowardWolosky: Regarding changes to the  `GitHubRepository` module being out-of-scope for this PR...

I think it's fine if you want to operate that way.  From my perspective, you're introducing an API that uses a new property from an existing object, so exposing that property for easy pipelining seems in-scope.  If you don't want to tackle that as part of this change, that's fine, but then I think we should get an issue opened to track having someone do that work as a follow-up.

GitHubTeams.ps1
HowardWolosky: Ah, just re-looked at the code and realized what I was saying and I stand by it.

This should have:
   * `ParentTeamName` - Just a normal parameter with no pipeline support
   * `ParentTeamSlug` (aliased to `TeamSlug`) - A parameter with `ValueFromPipelineByPropertyname`.  This way, someone can pipe in a parent team and have it be the parent and avoid the extra call needed to resolve the parent team by name.

HowardWolosky: Or actually, looking at the code again more speficially, `ParentTeamId` (aliased to `[TeamId`]).  But, logic otherwise follows suit.

GitHubTeams.ps1
```diff
@@ -276,6 +313,484 @@ filter Get-GitHubTeamMember
...
+        $getGitHubTeamParms['AccessToken'] = $AccessToken
+    }
+    if ($PSBoundParameters.ContainsKey('NoStatus'))
+    {
+        $getGitHubTeamParms['NoStatus'] = $NoStatus
+    }
+
+    $team = Get-GitHubTeam @getGitHubTeamParms
+
+    $uriFragment = "/orgs/$OrganizationName/teams/$($team.slug)"
```
HowardWolosky, line 768: Similar feedback for New/Set.  We should enable the ability to call this function without the need for the additional Get to retrieve the slug.  If a user pipes in an existing team to this function, it should have all the information it needs to be able to delete it.  That means that the Slug needs to be an alternative parameterset for this function.

GitHubTeams.ps1
```diff
@@ -276,6 +313,484 @@ filter Get-GitHubTeamMember
...
+    )
+
+    Write-InvocationLog
+
+    $telemetryProperties = @{
+        OrganizationName = (Get-PiiSafeString -PlainText $OrganizationName)
+        TeamName = (Get-PiiSafeString -PlainText $TeamName)
+    }
+
+    $getGitHubTeamParms = @{
```
HowardWolosky, line 599: Similar to New/Remove, there should be a parameterset that doesn't require there to be an ancillary `Get` call just to complete the normal case.  That means that the `TeamSlug` needs to be an alternative parameterset to `TeamName` and it should support pipeline input by property name so that it can be piped in.

GitHubTeams.ps1
```diff
@@ -276,6 +313,484 @@ filter Get-GitHubTeamMember
...
+        [string] $AccessToken,
+
+        [switch] $NoStatus
+    )
+
+    Write-InvocationLog
+
+    $telemetryProperties = @{
+        OrganizationName = (Get-PiiSafeString -PlainText $OrganizationName)
+        TeamName = (Get-PiiSafeString -PlainText $TeamName)
```
HowardWolosky, line 596: Given the guidance below, it would be helpful to track the ParameterSetName (whether the more common case is to pass in the Name or the Slug.

GitHubTeams.ps1
```diff
@@ -276,6 +313,484 @@ filter Get-GitHubTeamMember
...
+        [switch] $Force,
+
+        [string] $AccessToken,
+
+        [switch] $NoStatus
+    )
+
+    $telemetryProperties = @{
+        OrganizationName = (Get-PiiSafeString -PlainText $RepositoryName)
+        TeamName = (Get-PiiSafeString -PlainText $TeamName)
```
HowardWolosky, line 748: Given the guidance below, it would be helpful to track the ParameterSetName (whether the more common case is to pass in the Name or the Slug.

GitHubTeams.ps1
X-Guardian: But then you would get the option of 'TeamSlug' for parameter completion of the function, which would actually be the ParentTeam. Pretty ugly.

GitHubTeams.ps1
X-Guardian: I really don't see why. They user of the function has no knowledge of what a `TeamSlug` is. What is the problem with the function resolving the name for them?

GitHubTeams.ps1
X-Guardian: I don't intend to implement slugs in this PR.

GitHubTeams.ps1
HowardWolosky: I ended up [explaining this](https://github.com/microsoft/PowerShellForGitHub/pull/238#discussion_r443268865) in your #238 PR.  Repeating it here though:

The Microsoft org has **2100 teams** (and that number continues to grow).

* Getting all of them with `Get-GitHubTeam -OrganizationName microsoft -NoStatus` adds an additional **22 seconds** to method execution.
* Calling it _with_ status (e.g.  `Get-GitHubTeam -OrganizationName microsoft`)  adds an additional **104** seconds.
* _Allowing_ users to call into the method providing the slug for the team would add an additional **0 seconds**. 

I agree that exposing the existence of slugs makes the signature a bit uglier, however _not_ exposing the option of using slugs means that the API call can needlessly take additional _minutes_ to complete.

GitHubTeams.ps1
HowardWolosky: It is certainly your prerogative to stand by that, however if you don't implement slugs in this PR, be prepared for the function signature to change in a follow-up PR (either by me or someone else) in order to fix this functionality gap.

GitHubTeams.ps1
X-Guardian: So what this tells me is first, disable by default that horrendous code which is Status, and second, look at the inefficiencies of `Invoke-GHRestMethod` and `Invoke-GHRestMethodMultipleResult`, performing multiple update checks, telemetry calls etc and maybe you will get that time down to less than 5 seconds.

HowardWolosky: Let's just put the status discussion to the side for the moment and focus on the `-NoStatus` time.

* `Invoke-UpdateCheck` takes on average **2 ms**.  (Remember, all it's doing after the first check for the day is a call to `Get-Date` and a single `if` comparison.  The added load to normal module functionality is minimal at best as the average run time demonstrates.

* Disabling telemetry shaved off **1 second** on average from the total.  That's not the silver bullet.  That being said, there is food for thought there to determine if enough value is being gained from having a telemetry event for each `Invoke-GHRestMethod` within the confines of a `Invoke-GHRestMethodMultipleResult`.  I'll need to consider that one a bit more separately.

They key thing to keep in mind here is that to retrieve 2100 teams it has to make **70** REST requests.

Let's look at this totally bare bones:

```powershell
Measure-Command {
      1..70 |% { 
          $null = Invoke-WebRequest -Headers $headers -UseDefaultCredentials -UseBasicParsing -Method Get -Uri 'https://api.github.com/organizations/$orgId/teams?page=$_'
      }
}
```

So, using pure basic PowerShell with nothing from this module takes **17.5 seconds** on average.  Which is **4.5 seconds** faster than what the module is doing.  Those extra few seconds are buying us:
  * Telemetry
  * Smart Objects (so, actual DateTime objects instead of date strings)
  * Additional properties and type names being added
  * Logging
  * Maybe some other stuff that I'm missing.

I'm not seeing any way that this call could get down to 5 seconds, let alone under 17.5 seconds.

I'm also not seeing real inefficiencies that `Invoke-GHRestMethod` is causing in the normal case.  If you are, then please point them out so that they can get fixed.

I do continue to see the opportunity for this call to be avoided entirely however in the scenario where someone chooses to pipe in the team object instead.  If they choose to pass in just a TeamName, then they will have to live with a longer execution time due to the reality of the need to get the slug (it would be no different than having done the same work that you're doing, but external to this function and then piping in the result of a `Where-Object`)...but if the user has the slug, I still see no reason to not give them the opportunity for faster execution and avoid the `Get`.

HowardWolosky: I did a little more testing for completeness:

```powershell
Set-GitHubConfiguration -SessionOnly -DefaultNoStatus -DisableLogging -DisableSmarterObjects -DisablePipelineSupport

Measure-Command { $null = Get-GitHubTeam -OrganizationName microsoft }
```

Came in, on average around **18.5 seconds**.  Adding in `-DisableTelemetry` reduced it by another **.5 second** or so.

So, `Invoke-GHRestMethod` is able to match the performance of raw `Invoke-WebRequest` in this scenario within around .5 second overall with no code change.  There might be some limited opportunities to make it slightly faster, but I don't believe that there's much in the way of negative module stuff getting in the way here.  Each of those things that were disabled in that test have demonstrable value, _and_ each of them has a clearly documented way to be disabled in the event that a user values slight performance gains in lieu of the added benefits those disabled features would have provided.

GitHubTeams.ps1
X-Guardian: So just to clarify the conclusion to these conversations, I I'm not intending to implement slug support in this PR as I have no interest in the functionality. In my view it can be added later by someone else through another PR as a non-breaking change.

USAGE.md
```diff
@@ -426,6 +435,55 @@ Get-GitHubUser
 
 ----------
 
+### Teams
+
+#### Getting teams in an Organization
+
+```powershell
+Get-GitHubTeam -OrganizationName Microsoft
```
HowardWolosky, line 443: Can you lowercase the 'microsoft' examples here?

GitHubTeams.ps1
```diff
@@ -317,6 +832,25 @@ filter Add-GitHubTeamAdditionalProperties
...
+                if ($item.html_url -match "^https?://$hostName/orgs/([^/]+)/?([^/]+)?(?:/.*)?$")
+                {
+                    $organizationName = $Matches[1]
+                }
+                else
+                {
+                    $organizationName = ''
+                }
+            }
+            Add-Member -InputObject $item -Name 'OrganizationName' -value $organizationName -MemberType NoteProperty -Force
```
HowardWolosky, line 852: nit: line feed

GitHubTeams.ps1
```diff
@@ -317,6 +832,25 @@ filter Add-GitHubTeamAdditionalProperties
...
+                $hostName = $(Get-GitHubConfiguration -Name 'ApiHostName')
+
+                if ($item.html_url -match "^https?://$hostName/orgs/([^/]+)/?([^/]+)?(?:/.*)?$")
+                {
+                    $organizationName = $Matches[1]
+                }
+                else
+                {
+                    $organizationName = ''
+                }
```
HowardWolosky, line 850: Simplification of regex and expression:

```suggestion
                $organizationName = [String]::Empty
                if ($item.html_url -match "^https?://$hostName/orgs/([^/]+)/.*$")
                {
                    $organizationName = $Matches[1]
                }                
```

GitHubTeams.ps1
```diff
@@ -276,6 +313,484 @@ filter Get-GitHubTeamMember
...
+
+    .PARAMETER NoStatus
+        If this switch is specified, long-running commands will run on the main thread
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .INPUTS
+        GitHub.Team
+        System.String
```
HowardWolosky, line 537: All of the pipeline inputs here are `ValueFromPipelineByPropertyName`.  `System.String` doesn't really make sense here (I don't think) unless you were accepting `ValueFromPipeline` on a string parameter.

GitHubTeams.ps1
```diff
@@ -276,6 +313,484 @@ filter Get-GitHubTeamMember
...
+        }
+        if ($PSBoundParameters.ContainsKey('AccessToken'))
+        {
+            $getGitHubTeamParms['AccessToken'] = $AccessToken
+        }
+        if ($PSBoundParameters.ContainsKey('NoStatus'))
+        {
+            $getGitHubTeamParms['NoStatus'] = $NoStatus
+        }
+        $team = Get-GitHubTeam @getGitHubTeamParms
```
HowardWolosky, line 640: nit: line feed

GitHubTeams.ps1
```diff
@@ -276,6 +313,484 @@ filter Get-GitHubTeamMember
...
+        If provided, this will be used as the AccessToken for authentication with the
+        REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
+
+    .PARAMETER NoStatus
+        If this switch is specified, long-running commands will run on the main thread
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .OUTPUTS
```
HowardWolosky, line 690: `.INPUTS` ?

GitHubTeams.ps1
```diff
@@ -276,6 +313,484 @@ filter Get-GitHubTeamMember
...
+        If this switch is specified, long-running commands will run on the main thread
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .INPUTS
+        GitHub.Team
+        System.String
+
+    .OUTPUTS
```
HowardWolosky, line 539: This `.OUTPUTS` doesn't look right.

GitHubTeams.ps1
```diff
@@ -153,8 +161,37 @@ filter Get-GitHubTeam
...
-        Add-GitHubTeamAdditionalProperties)
+    $result = Invoke-GHRestMethodMultipleResult @params | Add-GitHubTeamAdditionalProperties
+
+    if ($PSBoundParameters.ContainsKey('TeamName'))
+    {
+        $team = $result | Where-Object -Property name -eq $TeamName
+
+        if ($null -eq $team)
+        {
+            throw "Team '$TeamName' not found"
```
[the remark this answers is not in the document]
X-Guardian: Done

GitHubTeams.ps1
X-Guardian: `System.String` removed.

GitHubTeams.ps1
X-Guardian: Fixed

GitHubTeams.ps1
X-Guardian: Done

GitHubTeams.ps1
X-Guardian: Added

GitHubTeams.ps1
X-Guardian: Done

GitHubTeams.ps1
X-Guardian: Done

USAGE.md
X-Guardian: Done

GitHubTeams.ps1
```diff
@@ -153,8 +161,37 @@ filter Get-GitHubTeam
...
+        }
+        else
+        {
+            $uriFragment = "/teams/$($team.id)"
```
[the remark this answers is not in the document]
X-Guardian: Yep, this needs changing to the new API. I didn't add this code in this PR, only changed the indent, so best if it is updated in another PR.

GitHubTeams.ps1
X-Guardian: This would need adding in another PR.

GitHubTeams.ps1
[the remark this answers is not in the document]
X-Guardian: How about `GitHub.Team` and `GitHub.TeamDetail`? Would you want this implemented as two separate `Add-*AdditionalProperties` functions,or just use the current one and override the TypeName?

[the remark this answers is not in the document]
HowardWolosky: `GitHub.TeamDetail` sounds perfect.
You can just use the current one and override `TypeName`.  You should probably update `Add-GitHubTeamAdditionalProperties` to add the additional properties to `$item.organization` if it's non-null.

GitHubTeams.ps1
HowardWolosky: sounds good.  I'll handle the slug update once this goes in.  And your feedback around the status handling from a different thread has been heard (and I agree with you), and my current thoughts are captured in #247.  I'd be interested in your thoughts there.

X-Guardian:
@HowardWolosky, do you want me to close this PR and open another one?

HowardWolosky:
Yes, please do.

X-Guardian:
This PR has been superseded by PR #257

PR #238, opened 2020-06-14
New GitHubRepositoryBranchProtectionRule Functions

X-Guardian:
#### Description

This PR adds the following GitHub Branch Protection Rule functions to the `GitHubBranches` module:

- `Get-GitHubRepositoryBranchProtectionRule`
- `Set-GitHubRepositoryBranchProtectionRule`
- `Remove-GitHubRepositoryBranchProtectionRule`

The `lukeCageAcceptHeader` was also added to the `GitHubCore` module with a value of `application/vnd.github.luke-cage-preview+json`.

#### Issues Fixed

None

##### GitHub REST API v3

- [Get Branch Protection](https://developer.github.com/v3/repos/branches/#get-branch-protection)
- [Update Branch Protection](https://developer.github.com/v3/repos/branches/#update-branch-protection)
- [Delete Branch Protection](https://developer.github.com/v3/repos/branches/#delete-branch-protection)

#### Checklist


X-Guardian:

GitHubBranches.ps1
```diff
@@ -154,6 +154,518 @@ filter Get-GitHubRepositoryBranch
...
+
+    .PARAMETER RepositoryName
+        Name of the repository.
+        If not supplied here, the DefaultRepositoryName configuration property value will be used.
+
+    .PARAMETER StatusChecks
+        The list of status checks to require in order to merge into the branch.
+
+    .PARAMETER RequireUpToDateBranches
+        Require branches to be up to date before merging.
```
HowardWolosky, line 294:
```suggestion
        Require branches to be up to date before merging.  This setting will not take effect unless at least one status check is defined.
```

GitHubBranches.ps1
```diff
@@ -154,6 +154,518 @@ filter Get-GitHubRepositoryBranch
...
+        {
+            $RestrictPushUsers = @()
+        }
+
+        if ($null -eq $RestrictPushTeams)
+        {
+            $RestrictPushTeams = @()
+        }
+
+        $restrictions = @{
```
HowardWolosky, line 486: Similar to my feedback above regarding `required_status_checks`, this will introduce regressions if users try to update _just_ the push users or _just_ the teams.  If I already have values defined for users, and now I go and want to make a separate call to set the teams, this code, as written, will wipe out my defined users.

GitHubBranches.ps1
```diff
@@ -107,3 +107,514 @@ function Get-GitHubRepositoryBranch
...
+
+    .PARAMETER RequiredApprovingReviewCount
+        Specify the number of reviewers required to approve pull requests. Use a number between 1
+        and 6.
+
+    .PARAMETER RestrictPushUsers
+        Specify which users have push access.
+
+    .PARAMETER RestrictPushTeams
+        Specify which teams have push access.
```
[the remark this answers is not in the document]
HowardWolosky: Actually, now I see that you're explicitly looking up the slugs inside the function and are expecting the team names here.  But that's unnecessary work if the user happened to pipe in team objects (if you choose to support that).

GitHubBranches.ps1
```diff
@@ -154,6 +154,518 @@ filter Get-GitHubRepositoryBranch
...
+        }
+
+        if ($null -eq $RestrictPushTeams)
+        {
+            $RestrictPushTeams = @()
+        }
+
+        $restrictions = @{
+            users = $RestrictPushUsers
+            teams = $RestrictPushTeams
```
HowardWolosky, line 488: This is inconsistent.  You call `Get-GitHubTeam` to resolve team names -> slugs for `DismissalTeams` but not for `RestrictPushTeams`?

GitHubBranches.ps1
```diff
@@ -154,6 +154,518 @@ filter Get-GitHubRepositoryBranch
...
+        if ($PSBoundParameters.ContainsKey('RestrictPushApps')) {
+            $restrictions['apps'] = $RestrictPushApps }
+    }
+    else
+    {
+        $restrictions = $null
+    }
+
+    $hashBody = @{
+        name = $RepositoryName
```
HowardWolosky, line 500: I don't see this being part of the spec for the request body

HowardWolosky:
@X-Guardian - please be sure to @ mention me once this is ready for the next review (there are so many of your reviews for me to keep track of).  Thanks again for all your work.  Much appreciated!

GitHubBranches.ps1
```diff
@@ -107,3 +107,514 @@ function Get-GitHubRepositoryBranch
...
+        $hashBody['allow_deletions'] = $AllowDeletions.ToBool() }
+
+    if ($PSCmdlet.ShouldProcess("'$Name' branch of repository '$RepositoryName'",
+        'Set GitHub Repository Branch Protection'))
+    {
+        Write-InvocationLog
+
+        $params = @{
+            UriFragment = "repos/$OwnerName/$RepositoryName/branches/$Name/protection"
+            Body = (ConvertTo-Json -InputObject $hashBody -Depth 3)
```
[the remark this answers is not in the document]
X-Guardian: Done, but not script level and only to 3.

GitHubBranches.ps1
X-Guardian: Removed

GitHubBranches.ps1
X-Guardian: Code added to support team names.

GitHubBranches.ps1
X-Guardian: GitHub have really over-complicated this API and it doesn't allow you to do this. To achieve this functionality, all the other individual API endpoints would need to be implemented. I was originally thinking of making this function `New-GitHubRepositoryBranchProtectionRule` and am happy to go back to that if required.

GitHubBranches.ps1
```diff
@@ -107,3 +107,514 @@ function Get-GitHubRepositoryBranch
...
+
+    $telemetryProperties = @{
+        OwnerName = (Get-PiiSafeString -PlainText $OwnerName)
+        RepositoryName = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    if ($PSBoundParameters.ContainsKey('StatusChecks'))
+    {
+        $requiredStatusChecks = @{
+            strict = $RequireUpToDateBranches.ToBool()
```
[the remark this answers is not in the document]
X-Guardian: See comment below.

GitHubBranches.ps1
```diff
@@ -107,3 +107,514 @@ function Get-GitHubRepositoryBranch
...
+        and 6.
+
+    .PARAMETER RestrictPushUsers
+        Specify which users have push access.
+
+    .PARAMETER RestrictPushTeams
+        Specify which teams have push access.
+
+    .PARAMETER RestrictPushApps
+        Specify which apps have push access.
```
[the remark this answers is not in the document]
X-Guardian: I really don't think user of the module should need to know about slugs. We don't currently have a `Get-GitHubApp` function that would be needed for this, so I'll raise another PR to add that, and then this function can use that to resolve the App name to a slug internally.

GitHubBranches.ps1
```diff
@@ -107,3 +107,514 @@ function Get-GitHubRepositoryBranch
...
+        The list of status checks to require in order to merge into the branch.
+
+    .PARAMETER RequireUpToDateBranches
+        Require branches to be up to date before merging.
+
+    .PARAMETER EnforceAdmins
+        Enforce all configured restrictions for administrators.
+
+    .PARAMETER DismissalUsers
+        Specify which users can dismiss pull request reviews.
```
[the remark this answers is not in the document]
X-Guardian: Done

GitHubBranches.ps1
```diff
@@ -107,3 +107,514 @@ function Get-GitHubRepositoryBranch
...
+    $RepositoryName = $elements.repositoryName
+
+    $telemetryProperties = @{
+        OwnerName = (Get-PiiSafeString -PlainText $OwnerName)
+        RepositoryName = (Get-PiiSafeString -PlainText $RepositoryName)
+    }
+
+    if ($PSBoundParameters.ContainsKey('StatusChecks'))
+    {
+        $requiredStatusChecks = @{
```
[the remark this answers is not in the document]
X-Guardian: Don't understand. it is an if/else block. that sets `requiredStatusChecks` for both paths.

GitHubBranches.ps1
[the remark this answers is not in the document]
HowardWolosky: The Microsoft org has 2100 teams.

* Getting all of them with `Get-GitHubTeam -OrganizationName microsoft -NoStatus` adds an additional **22 seconds** to method execution.
* Calling it _with_ status adds an additional **104** seconds.
* Allowing users to call into the method providing the slugs for the teams would add an additional **0 seconds**. 

**Double** that time if they happen to pass in both Dismissal and Restriction teams, since you do it _twice_.  (By the way, you should also optimize the code to only call `Get-GitHubTeam` _once_ if any sort of team is provided).

I agree that exposing the existence of slugs makes the signature a bit uglier, however _not_ exposing the option of using slugs means that the API call can needlessly take additional _minutes_ to complete.

GitHubBranches.ps1
```diff
@@ -154,6 +155,549 @@ filter Get-GitHubRepositoryBranch
...
+    {
+        $hashBody['allow_force_pushes'] = $AllowForcePushes.ToBool()
+    }
+    if ($PSBoundParameters.ContainsKey('AllowDeletions'))
+    {
+        $hashBody['allow_deletions'] = $AllowDeletions.ToBool()
+    }
+
+    if ($PSCmdlet.ShouldProcess("'$BranchName' branch of repository '$RepositoryName'",
+        'Set GitHub Repository Branch Protection'))
```
HowardWolosky, line 548:
```suggestion
    if ($PSCmdlet.ShouldProcess(
        "'$BranchName' branch of repository '$RepositoryName'",
        'Set GitHub Repository Branch Protection'))
```

GitHubBranches.ps1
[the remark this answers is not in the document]
HowardWolosky: And actually, saying that and then reviewing the code, it occurs to me that you don't have `OrganizationName` as a parameter, which means that users can't specify team names that are owned by an organization, which could be a problem if the repo that the rule is being applied to is one owned by an organization.

GitHubBranches.ps1
HowardWolosky: I don't follow.  You've already implemented `Get-GitHubRepositoryBranchProtectionRule`.  If you detected that any of these object properties were being altered and called that first, wouldn't you have all the content that you'd need in order to ensure you weren't overwriting any property that the user didn't provide in this call?

GitHubBranches.ps1
```diff
@@ -202,3 +746,64 @@ filter Add-GitHubBranchAdditionalProperties
...
+
+        if (-not (Get-GitHubConfiguration -Name DisablePipelineSupport))
+        {
+            $elements = Split-GitHubUri -Uri $item.url
+            $repositoryUrl = Join-GitHubUri @elements
+            Add-Member -InputObject $item -Name 'RepositoryUrl' -Value $repositoryUrl -MemberType NoteProperty -Force
+
+            $hostName = $(Get-GitHubConfiguration -Name 'ApiHostName')
+
+            if ($item.html_url -match "^https?://api.$hostName/repos/([^/]+)/?([^/]+)/?([^/]+)/?([^/]+)?(?:/.*)?$")
```
HowardWolosky, line 795: This failed to get `master` from `https://api.github.com/repos/microsoft/PowerShellForGitHub/branches/master/protection`


```suggestion
            $hostName = $(Get-GitHubConfiguration -Name 'ApiHostName')
            if ($item.html_url -match "^https?://(?:www\.|api\.|)$hostName/repos/(?:[^/]+)/(?:[^/]+)/branches/([^/]+)/.*$")
```

And then this would match the branch name against `$Matches[1]`

GitHubBranches.ps1
```diff
@@ -202,3 +746,64 @@ filter Add-GitHubBranchAdditionalProperties
...
+    param(
+        [Parameter(
+            Mandatory,
+            ValueFromPipeline)]
+        [AllowNull()]
+        [AllowEmptyCollection()]
+        [PSCustomObject[]] $InputObject,
+
+        [ValidateNotNullOrEmpty()]
+        [string] $TypeName = $script:GitHubBranchTypeName
```
HowardWolosky, line 780:
```suggestion
        [string] $TypeName = $script:GitHubBranchProtectionRuleName
```

GitHubBranches.ps1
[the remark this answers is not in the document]
HowardWolosky: I would define `$requiredStatusChecks = $null` outside of the `if`, and then you won't need the `else` at all.
We don't currently have strict mode on right now, but I still prefer to code that way.  The way you have it right now, you're declaring a variable within the `if` scope, so while PowerShell allows it to be accessed outside of that scope, it's still not a great practice to follow.  You generally want to be declaring/initially using your variables in the scope you're intending to use them, which in thi scase, is outside of the `if` scope.

X-Guardian:
This PR is superseded by PR #255

PR #241, opened 2020-06-16
GitHubContents: New Function Set-GitHubContent

X-Guardian:
#### Description

This PR adds the following new function to the `GitHubContents` module:

- `Set-GitHubContent`

An additional parameter `BranchName` has also been added to the `Get-GitHubContent` function which is required by `Set-GitHubContent`.

#### Issues Fixed

None

#### References

[GitHub Rest API v3 Contents](https://developer.github.com/v3/repos/contents/#create-or-update-file-contents)

#### Checklist


X-Guardian:

GitHubContents.ps1
```diff
@@ -78,6 +81,8 @@ function Get-GitHubContent
 
         [string] $Path,
 
+        [string] $BranchName,
```
HowardWolosky, line 84:
```suggestion
        [ValidateNotNullOrEmpty()]
        [string] $BranchName,
```

GitHubContents.ps1
```diff
@@ -114,6 +119,11 @@ function Get-GitHubContent
         $description = "Getting all content for in $RepositoryName"
     }
 
+    if ($PSBoundParameters.ContainsKey('Branch'))
+    {
+        $uriFragment += "?ref=$Branch"
```
HowardWolosky, line 124: Wrong name

```suggestion
    if ($PSBoundParameters.ContainsKey('BranchName'))
    {
        $uriFragment += "?ref=$BranchName"
```

GitHubContents.ps1
```diff
@@ -143,3 +153,266 @@ function Get-GitHubContent
...
+            Mandatory,
+            Position = 2)]
+        [string] $CommitMessage,
+
+        [Parameter(
+            Mandatory,
+            Position = 3)]
+        [string] $Content,
+
+        [Parameter(
```
HowardWolosky, line 246: I disagree with the ordering here of the parameters.  I've mentioned this in other PR's today as well.  I feel that the repo being operated on is the source, and should be the initial parameter.  It's consistent with the rest of the repo, and it's consistent with how other built-in PowerShell functions work.

GitHubContents.ps1
```diff
@@ -143,3 +153,266 @@ function Get-GitHubContent
...
+        REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
+
+    .PARAMETER NoStatus
+        If this switch is specified, long-running commands will run on the main thread
+        with no commandline status update.  When not specified, those commands run in
+        the background, enabling the command prompt to provide status information.
+        If not supplied here, the DefaultNoStatus configuration property value will be used.
+
+    .EXAMPLE
+        Set-GitHubContent  -Path README.md -OwnerName microsoft -RepositoryName PowerShellForGitHub -CommitMessage 'Adding README.md' -Content '# README' -BranchName master
```
HowardWolosky, line 220: nit: extra space

```suggestion
        Set-GitHubContent -Path README.md -OwnerName microsoft -RepositoryName PowerShellForGitHub -CommitMessage 'Adding README.md' -Content '# README' -BranchName master
```

HowardWolosky: You'll also likely want to update the parameter ordering too once that gets changed.

GitHubContents.ps1
```diff
@@ -143,3 +153,266 @@ function Get-GitHubContent
...
+            ![System.String]::IsNullOrEmpty($CommitterEmail))
+        {
+            $hashBody['committer'] = @{
+                name = $CommitterName
+                email = $CommitterEmail
+            }
+        }
+        else
+        {
+            throw 'Both CommiterName and CommitterEmail need to be specified.'
```
HowardWolosky, line 312:
```suggestion
            $message = 'Both CommiterName and CommitterEmail need to be specified.'
            Write-Log -Message $message -Level Error
            throw $message
```

Here and below

GitHubContents.ps1
```diff
@@ -143,3 +153,266 @@ function Get-GitHubContent
...
+        The OwnerName and RepositoryName will be extracted from here instead of needing to provide
+        them individually.
+
+    .PARAMETER Path
+        The file path for which to set contents.
+
+    .PARAMETER CommitMessage
+        The Git commit message.
+
+    .PARAMETER Content
```
HowardWolosky, line 187: It seems that you might want to support two parameter sets...one where the user specifies the `Content` directly, and the other where they specify a `ContentPath` to a file that has the content that they want to use.

GitHubContents.ps1
```diff
@@ -143,3 +153,266 @@ function Get-GitHubContent
 
     return $result
 }
+
+function Set-GitHubContent
```
HowardWolosky, line 157: Please update to be a `filter` with pipeline support.

GitHubContents.ps1
```diff
@@ -143,3 +153,266 @@ function Get-GitHubContent
...
+                }
+
+                if ($PSBoundParameters.ContainsKey('AccessToken'))
+                {
+                    $getGitHubContentParms['AccessToken'] = $AccessToken
+                }
+
+                if ($PSBoundParameters.ContainsKey('NoStatus'))
+                {
+                    $getGitHubContentParms['NoStatus'] = $NoStatus
```
HowardWolosky, line 406: This won't work right for anyone with `DefaultNoStatus:$true` (like Pester tests).  You'll need to use `Resolve-ParameterWithDefaultConfigurationValue -Name NoStatus -ConfigValueName DefaultNoStatus`

GitHubContents.ps1
```diff
@@ -143,3 +153,266 @@ function Get-GitHubContent
...
+                -ConfigValueName DefaultNoStatus)
+        }
+
+        try
+        {
+            return Invoke-GHRestMethod @params
+        }
+        catch
+        {
+            $overwriteShaRequired = $false
```
HowardWolosky, line 356: This is all well and good, but I think you should still explicitly expose a `Sha` parameter.  That way this extra check can be avoided if a user pipes in a `GitHub.Content` object which already has the sha and all of the other info).

GitHubContents.ps1
```diff
@@ -143,3 +153,266 @@ function Get-GitHubContent
...
+                }
+                else
+                {
+                    throw $_
+                }
+            }
+
+            if ($overwriteShaRequired)
+            {
+                # Get SHA from current file
```
HowardWolosky, line 387: This scenario warrants a verbose log entry to explain what's going on...something like:
```powershell
Write-Log -Level Verbose -Message 'Replacing the content of an existing file requires the current SHA of that file.  Retrieving the SHA now.'
```

Tests/GitHubContents.tests.ps1
```diff
@@ -172,6 +179,176 @@ try
...
+                $setGitHubContentParms = @{
+                    Path = "$filePath/$fileName"
+                    CommitMessage = $commitMessage
+                    BranchName = $branchName
+                    Content = $content
+                    Uri = $repo.svn_url
+                    CommitterName = $committerName
+                    CommitterEmail = $committerEmail
+                    AuthorName = $authorName
+                    authorEmail = $authorEmail
```
HowardWolosky, line 266:
```suggestion
                    AuthorEmail = $authorEmail
```

Tests/GitHubContents.tests.ps1
```diff
@@ -172,6 +179,176 @@ try
 
         Remove-GitHubRepository -Uri $repo.svn_url -Confirm:$false
     }
+
+    Describe 'GitHubContents/Set-GitHubContent' {
```
HowardWolosky, line 183: I **really** appreciate the awesome and thorough UT's that you've been writing for all of your submissions.  SO, so much appreciated.  Thank you.

GitHubContents.ps1
```diff
@@ -143,3 +153,266 @@ function Get-GitHubContent
...
+            }
+        }
+        else
+        {
+            throw 'Both AuthorName and AuthorEmail need to be specified.'
+        }
+    }
+
+    if ($PSCmdlet.ShouldProcess("$BranchName branch of $RepositoryName",
+        "Set GitHub Contents on $Path"))
```
HowardWolosky, line 334: Generally, my preference when splitting a parameter-supplied function call across multiple lines (regardless of language) is to put _each_ parameter on its own line:

```suggestion
    if ($PSCmdlet.ShouldProcess(
        "$BranchName branch of $RepositoryName",
        "Set GitHub Contents on $Path"))
```

HowardWolosky:
@X-Guardian - please be sure to @ mention me once this is ready for the next review.  Thanks again for all your work.  Much appreciated!

GitHubContents.ps1
X-Guardian: Done

X-Guardian:

GitHubContents.ps1
```diff
@@ -311,6 +341,10 @@ function Set-GitHubContent
             ParameterSetName = 'Elements')]
         [string] $RepositoryName,
```
HowardWolosky, line 342: I prefer for the parameters to be in the same order as their CBH entries.  `OwnerName`/`RepositoryName` should really be at the top before `Uri`.  I'm ok with the fact that they'd be listed before a parameter with `Position = 1`.

GitHubContents.ps1
```diff
@@ -269,37 +275,61 @@ function Set-GitHubContent
...
 
         Sets the contents of the README.md file on the master branch of the PowerShellForGithub repository.
 #>
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', '',
+        Justification = 'One or more parameters (like NoStatus) are only referenced by helper
+        methods which get access to it from the stack via Get-Variable -Scope 1.')]
     [CmdletBinding(
         SupportsShouldProcess,
         PositionalBinding = $false,
         DefaultParameterSetName = 'Elements')]
```
HowardWolosky, line 307: Looks like you need to add a test involving the pipeline.  If you try to pipe in the repo to this, it still wants `OwnerName`/`RepositoryName` due to this default that you have here.

HowardWolosky: I recommend an additional test for any parameter that can be piped in (that's what I ended up doing with the pipeline change which is why the test count exploded...but it caught a number of bugs along the way).

GitHubContents.ps1
```diff
@@ -508,11 +556,37 @@ filter Add-GitHubContentAdditionalProperties
...
+
+            if ($uri -match "^https?://$hostName/?([^/]+)/?([^/]+)/blob/?([^/]+)/?([^#]*)?$")
+            {
+                $branchName = $Matches[3]
+                $path = $Matches[4]
+            }
+            else
+            {
+                $branchName = ''
+                $path = ''
```
HowardWolosky, line 583:
```suggestion
                $branchName = [String]::Empty
                $path = [String]::Empty
```

GitHubContents.ps1
```diff
@@ -508,11 +556,37 @@ filter Add-GitHubContentAdditionalProperties
...
+            }
+
+            $elements = Split-GitHubUri -Uri $uri
             $repositoryUrl = Join-GitHubUri @elements
+
             Add-Member -InputObject $item -Name 'RepositoryUrl' -Value $repositoryUrl -MemberType NoteProperty -Force
+
+            $hostName = $(Get-GitHubConfiguration -Name 'ApiHostName')
+
+            if ($uri -match "^https?://$hostName/?([^/]+)/?([^/]+)/blob/?([^/]+)/?([^#]*)?$")
```
HowardWolosky, line 575:
```suggestion
            $hostName = $(Get-GitHubConfiguration -Name 'ApiHostName')
            if ($uri -match "^https?://$hostName/([^/]+)/([^/]+)/(blob|tree)/([^/]+)/([^#]*)?$")
```

I simplified the RegEx a little bit, and made it work for both individual files as well as when a directory list is returned (we won't have a filename, but we'll still have a branch name).

GitHubContents.ps1
```diff
@@ -508,11 +556,37 @@ filter Add-GitHubContentAdditionalProperties
...
+            {
+                $branchName = $Matches[3]
+                $path = $Matches[4]
```
HowardWolosky, line 578: With the regex change above, you'll need to update these match references as well:

```suggestion
                $branchName = $Matches[4]
                $path = $Matches[5]
```

Tests/GitHubContents.tests.ps1
```diff
@@ -254,6 +254,172 @@ try
...
+            }
+
+            It 'Should have written the correct content' {
+                $getGitHubContentParms = @{
+                    Path = "$filePath/$fileName"
+                    Uri = $repo.svn_url
+                    MediaType = 'Raw'
+                    ResultAsString = $true
+                }
+                $writtenContent = Get-GitHubContent @getGitHubContentParms
```
HowardWolosky, line 366: nit: linefeed

Tests/GitHubContents.tests.ps1
```diff
@@ -254,6 +254,172 @@ try
...
+                    BranchName = $branchName
+                    Content = $content
+                    Uri = $repo.svn_url
+                    CommitterName = $committerName
+                    CommitterEmail = $committerEmail
+                    authorName = $authorName
+                    authorEmail = $authorEmail
+                    NoStatus = $true
+                }
+                $result = Set-GitHubContent @setGitHubContentParms
```
HowardWolosky, line 343: nit: line feed

GitHubContents.ps1
HowardWolosky: Actually, even better...specify non-capturing groups so that you can just always use `$Matches[1]` and `$Matches[2]`:

```powershell
            if ($uri -match "^https?://(?:www\.|api\.|)$hostName/(?:[^/]+)/(?:[^/]+)/(?:blob|tree)/([^/]+)/([^#]*)?$")
```

Tests/GitHubContents.tests.ps1
```diff
@@ -254,6 +254,174 @@ try
             }
         }
     }
+
+    Describe 'GitHubContents/Set-GitHubContent' {
```
HowardWolosky, line 258: I'd still like to see at least one test involving pipelining to validate that your removal of `DefaultParameterSetName` fixed the problem.

X-Guardian: Done

GitHubContents.ps1
```diff
@@ -235,11 +555,37 @@ filter Add-GitHubContentAdditionalProperties
...
+            $hostName = $(Get-GitHubConfiguration -Name 'ApiHostName')
+
+            if ($uri -match "^https?://(?:www\.|api\.|)$hostName/(?:[^/]+)/(?:[^/]+)/(?:blob|tree)/([^/]+)/([^#]*)?$")
+            {
+                $branchName = $Matches[1]
+                $path = $Matches[2]
+            }
+            else
+            {
+                $branchName = [String]::Empty
```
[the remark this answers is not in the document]
HowardWolosky: This is a pattern I stand by.  It's a good general practice to ensure that the variables that you're intending to use are initialized as expected at the scope that you're intending to use them.  I have seen unintentional consequences of not following this pattern in past projects where a local variable was being used and it happened that there was another variable at a higher scope with the same name that the implementer (and reviewer) didn't realize at the time.  In those cases, there didn't happen to be `else` that initialized the alternate case (as they had expected default `$null`/`0` behavior from uninitialized variables) which clearly didn't happen in certain scenarios due to the higher scope'd variable.  I admit that the code in this specific instance is robust to that problem, but it still sits on top of a [code smell](https://en.wikipedia.org/wiki/Code_smell) that I'd like to keep out of this project.

In other languages doing what you're doing would be strictly prohibited. I try to take the middle-ground of taking advantage of PowerShell benefits (like not needing to test if an object is `$null` before iterating over its members), while following learnings from past projects like simply ensuring that variables are initialized at the scope that that they'll be getting used.  That's why you'll see `$uriFragment` and `$description` often getting initialized at the top of functions where there is additional logic internally to determine their values:

https://github.com/microsoft/PowerShellForGitHub/blob/0bc09d28714eb5be522471dde25790946286bcdd/GitHubIssueComments.ps1#L213-L217

X-Guardian:
@HowardWolosky, can we trigger the CI for this PR?

X-Guardian:
Ok @HowardWolosky, ready for next review.

GitHubContents.ps1
```diff
@@ -197,6 +208,315 @@
...
+    .PARAMETER Content
+        The new file content.
+
+    .PARAMETER Sha
+        The SHA value of the current file if present. If this parameter is not provided, and the
+        file currently exists in the specified branch of the repo, it will be read to obtain this
+        value.
+
+    .PARAMETER BranchName
+        The branch, or defaults to the default branch of not specified.
```
HowardWolosky, line 250:
```suggestion
        The branch, or defaults to the default branch if not specified.
```

HowardWolosky: I'll fix this myself after mergine.

PR #242, opened 2020-06-17
Adding complete Pipeline support to the entire module

HowardWolosky:
# Description

Comprehensively adds pipeline support throughout the entire module

This change required examining every method, and in some cases a few breaking changes (noted below) had to be introduced in order to make support work consistently end-to-end.  UT's were added for every function, which resuled in a few bugs that were identified and fixed (also noted below).

### Unavoidable Breaks
 * `Get-GitHubRepositoryBranch`: `Name` parameter is now `BranchName`.
  > A `GitHub.Repository` object has a `name` property which is the name of the repo.  Piping in the repo object to this method would confuse it, making it search for a branch with the name of the repo, as opposed to the desired effect of just listing out all branches in the repo.

 * `*-GitHub*Label`: `Name` parameter is now `Label`.
  > Similar to above.
  > `Name` was too generic and was causing unintended conflicts when passing in certain objects over the pipeline (like a `GitHub.Repository`) which also has a `name` property, making it impossible to pipe in a repository to list all labels (since it was trying to query for a label in that repository named the same as the repo.

 * `*-GitHubLabel`: `Milestone` parameter is now `MilestoneNumber`.
  > A `GitHub.Issue` contains a `milestone` object property, and PowerShell was complaining that there was no way to convert that to an `[int64]` when an Issue was passed in via the pipeline.  The only way to avoid this was to use `MilestoneNumber` which is the name of the property we add to `GitHub.Milestone` objects, and it's what we alias to in all other methods that interact with milestones.

 * `*-GitHubIssue`: `Milestone` parameter is now `MilestoneNumber`.
  > A `GitHub.Issue` contains a `milestone` object property, and PowerShell was complaining that there was no way to convert that to an `[int64]` when an Issue was passed in via the pipeline.  The only way to avoid this was to use `MilestoneNumber` which is the name of the property we add to `GitHub.Milestone` objects, and it's what we alias to in all other methods that interact with milestones.

 * `Get-GitHubLicense`: `Name` parameter is now `Key`.
  > The `key` is what you can query against, but a License object has _both_ a `name` and a `key` property, which caused piped object queries to fail with the existing parameter name.
 
 * `Get-GitHubCodeOfConduct`: `Name` parameter is now `Key`.
  > Similar to above.
  > The `key` is what you can query against, but a Code of Conduct object has _both_ a `name` and a `key` property, which caused piped object queries to fail with the existing parameter name.
 
[...]

Issue #110, opened 2019-05-20
GitHub Release support


rjmholt:
I wrote some PowerShell to publish a GitHub release recently ([here](https://github.com/PowerShell/vscode-powershell/blob/a7401961967f50f0de0dc64a82718b1ceb9d8ec5/tools/GitHubTools.psm1#L232-L328)) and was wondering if you would like it? It needs changing to work with the conventions of this module, of course, but with some guidance I can do that.

Let me know if you're interested. I haven't exposed any other GitHub Release API support, but could relatively easily.

HowardWolosky:
Go for it.  Would love to have it.
It also might be worthwhile to see if [vscode-powershell](https://github.com/PowerShell/vscode-powershell) could depend on this module for the GitHub integration functionality, as opposed to maintaining two implementations trying to do the same thing, one more generally than the other...

rjmholt:
> It also might be worthwhile to see if vscode-powershell could depend on this module for the GitHub integration functionality, as opposed to maintaining two implementations trying to do the same thing, one more generally than the other...

Yes that's my thinking too

PR #254, opened 2020-06-29
PowerShellForGitHub: Improve and Standardise WhatIf/Confirm Processing

### LINKED ISSUE #225: PowerShellForGitHub: Suggested WhatIf/Confirm Processing Improvements
#### Issue Details

Since PR #213, the `ShouldProcess` (WhatIf/Confirm) processing in the module is better, but it still could be improved.
The `PowerShellForGuthub` functions are designed to be consumed by other scripts/functions,
so the norm in this case is for each function to only ever output a single `ShouldProcess` prompt.
Currently, most functions output at least two prompts, for example:

```powershell
PS> New-GitHubRepository -OrganizationName test -RepositoryName test -WhatIf
What if: Performing the operation "Output to File" on target "\\nas01\data\users\simon\documents\PowerShellForGitHub.log".
What if: Performing the operation "Invoke-WebRequest" on target "https://api.github.com/orgs/test/repos".
```

The `WhatIf` for the 'Output to File' should not be shown, and the second `WhatIf` has a non-useful operation and target. It should be: `What if: Performing the operation "Create Repository" on target "test"`.

Also, if a function is given an invalid set of parameters, for example:

```powershell
New-GitHubRepository -RepositoryName test -TeamId 1 -WhatIf
What if: Performing the operation "Output to File" on target "\\nas01\data\users\simon\documents\PowerShellForGitHub.log".
Exception: Z:\Users\Simon\Documents\GitHub\X-Guardian\PowerShellForGitHub\GitHubRepositories.ps1:144:9
Line |
 144 |          throw $message
     |          ~~~~~~~~~~~~~~
     | TeamId may only be specified when creating a repository under an organization.
```     

The `WhatIf` for the 'Output to File' should not be shown.

#### Suggested solution to the issue

- Move the `$PSCmdlet.ShouldProcess` check out of `Invoke-GHRestMethod` and add it to the calling functions with the correct operation and target (as has been done for the `ConfirmHighImpact` PR #174).
- Move the `Write-InvocationLog -Invocation $MyInvocation` line in each resource function to within the `$PSCmdlet.ShouldProcess` condition block. This will prevent the log being updated if `ShouldProcess` is $false.
- Add `-WhatIf:$false` to the `Out-File` Cmdlet call in the `Write-Log` function in the `Helpers` module. This will prevent the `Write-Log` function displaying `WhatIf`.

This would also have the additional benefit of being able to remove the `SuppressMessageAttribute` `PSShouldProcess` decorator from all the functions, as the `ShouldProcess` logic will now be happening within the function.

#### Requested Assignment

- If possible, I would like to fix this.

#### Operating System

```
OsOperatingSystemSKU : 48
OsArchitecture       : 64-bit
WindowsVersion       : 1903
OsLanguage           : en-GB
OsMuiLanguages       : {en-GB, en-US}
```

#### PowerShell Version

```
PSVersion                      7.0.1
GitCommitId                    7.0.1
OS                             Microsoft Windows 10.0.18362
```

#### Module Version

Running: 0.14.0
Installed: 0.14.0

X-Guardian:
#### Description

This PR improves and standardises the WhatIf/Confirm ('ShouldProcess') processing across all the functions.

##### Summary of changes

- Moved the `$PSCmdlet.ShouldProcess` check out of `Invoke-GHRestMethod` and added it to the calling functions with the correct operation and target.
- Added `WhatIf:$false` and `Confirm:$false` to the `Out-File` Cmdlet call in the `Write-Log` function in the Helpers module. This will prevent the `Write-Log` function displaying `WhatIf` and `Confirm` prompts.
- Removed `ShouldProcess` from non-state changing functions.
- Modified the current `ShouldProcess` conditions to use an early return.
- Modified the `Set-GitHubIssueLabel` function to remove `ConfirmImpact='High'` and set `ConfirmPreference` to 'Low' if no labels have been specified instead.
- Modified the `Update-GitHubRepository` function to remove `ConfirmImpact='High'` and set `ConfirmPreference` to 'Low' if the repo is being renamed instead.

#### Issues Fixed

- Fixes #225

#### References

N/A

#### Checklist

- [ ] ~~Comment-based help added/updated, including examples.~~
- [ ] ~~Changes to the manifest file follow the [manifest guidance](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#module-manifest).~~
- [ ] ~~Unit tests were added/updated and are all passing. See [testing guidelines](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#testing).  This includes making sure that all pipeline input variations have been covered.~~
- [ ] ~~Relevant usage examples have been added/updated in [USAGE.md](https://github.com/microsoft/PowerShellForGitHub/blob/master/USAGE.md).~~


GitHubPullRequests.ps1
```diff
@@ -397,6 +391,7 @@ filter New-GitHubPullRequest
     if ($PSBoundParameters.ContainsKey('Title'))
     {
         $description = "Creating pull request $Title in $RepositoryName"
+        $shouldProcessAction = "Create GitHub Pull Request $Title"
```
HowardWolosky, line 394:
```suggestion
        $shouldProcessAction = "Create GitHub Pull Request: $Title"
```

GitHubRepositories.ps1
```diff
@@ -1241,6 +1236,18 @@ filter Set-GitHubRepository
     if ($PSBoundParameters.ContainsKey('IsTemplate')) { $hashBody['is_template'] = $IsTemplate.ToBool() }
     if ($PSBoundParameters.ContainsKey('Archived')) { $hashBody['archived'] = $Archived.ToBool() }
 
+    if ($Force -and (-not $Confirm))
+    {
+        $ConfirmPreference = 'None'
+    }
+
+    if (-not $PSCmdlet.ShouldProcess($RepositoryName, $shouldProcessMessage))
```
HowardWolosky, line 1244: We need to talk through this change a bit more ([previous conversation](https://github.com/microsoft/PowerShellForGitHub/pull/242#discussion_r441736038)).

`Rename-GitHubRepository` used to require confirmation.  In #242, this was refactored to just internally use `Set-GitHubRepository` which could already support that scenario.  Given that renaming a repository is one of those actions that is pretty destructive, it makes sense to me that it should require confirmation.  But we should only be requiring confirmation if the repo is being renamed.  So, I don't think that `ShouldProcess` should _always_ be happening...only if the repo is being renamed...but if it's being renamed, it should require confirmation.

Let's discuss further here.

HowardWolosky: Scratch that.  I've just experimented with this change more, and now I understand it better.  It's working exactly the way I was desiring.

Brilliant.
Well done.

Helpers.ps1
```diff
@@ -384,7 +383,7 @@ function Write-Log
                 }
                 else
                 {
-                    $logFileMessage | Out-File -FilePath $Path -Append
+                    $logFileMessage | Out-File -FilePath $Path -Append -WhatIf:$false -Confirm:$false
```
HowardWolosky, line 386: Why do we need `$Confirm:$false`?

```suggestion
                    $logFileMessage | Out-File -FilePath $Path -Append -WhatIf:$false
```

HowardWolosky:
Also -- please let me know in what order you'd prefer your changes to go in...if you're wanting this one to go in _before_ your other changes or after.

GitHubLabels.ps1
```diff
@@ -1104,16 +1140,26 @@ function Set-GitHubIssueLabel
             'labels' = $labelNames
         }
 
+        $shouldProcessMessage = "Set GitHub Issue Label(s) $($Label -join ', ')"
+
+        if ([System.String]::IsNullOrEmpty($Label))
+        {
+            $ConfirmPreference = 'Low'
```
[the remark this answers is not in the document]
X-Guardian: You wanted a `Confirm` prompt to be shown if all labels were removed. (i.e. `$Label` is null or empty.) but not otherwise.

`ConfirmPreference` defaults to `High` and `ConfirmImpact` defaults to `Medium` and a `Confirm` prompt is only shown when the `ConfirmImpact` argument is equal to or greater than the value of the `ConfirmPreference` variable, so setting `ConfirmPreference` to `Low` or `Medium` will cause the `Confirm` prompt to be shown.

GitHubPullRequests.ps1
X-Guardian: Done

Helpers.ps1
X-Guardian: Because otherwise if `Confirm` has been specified in a calling function, `Out-File` will re-prompt. I tested this to be 100% sure.

GitHubConfiguration.ps1
```diff
@@ -910,15 +926,15 @@ function Set-GitHubAuthentication
     }
 
     $script:accessTokenCredential = $Credential
+
+    Write-InvocationLog
```
HowardWolosky, line 930: This should be right after the `$PSCmdlet.ShouldProcess()` check, like elsewhere.  Otherwise, we could potentially have the error from line 924 without the invocation line in the log.

GitHubConfiguration.ps1
```diff
@@ -953,28 +969,27 @@ function Clear-GitHubAuthentication
         [switch] $SessionOnly
     )
 
-    Write-InvocationLog
+    if (-not $PSCmdlet.ShouldProcess('GitHub Authentication', 'Clear'))
+    {
+        return
+    }
 
     Set-TelemetryEvent -EventName Clear-GitHubAuthentication
```
HowardWolosky, line 977: Just from a pattern consistency perspective, can you put this _after_ the `Write-InvocationLog` call?

GitHubProjectCards.ps1
```diff
@@ -604,6 +610,13 @@ filter Move-GitHubProjectCard
         $hashBody.add('column_id', $Column)
     }
 
+    if (-not $PSCmdlet.ShouldProcess($Card, 'Move GitHub Project Card'))
+    {
+        return
+    }
+
+    Write-InvocationLog
```
HowardWolosky, line 618: A problem with this pattern (of putting `Write-InvocationLog` _after_ the `$PSCmdlet.ShouldProcess` is that error messages written to the log (like at line 588) won't have the invocation attribution to go along with it.

GitHubProjectColumns.ps1
```diff
@@ -473,6 +477,13 @@ filter Move-GitHubProjectColumn
         'position' = $Position
     }
 
+    if (-not $PSCmdlet.ShouldProcess($Column, 'Move GitHub Project Column'))
+    {
+        return
+    }
+
+    Write-InvocationLog
```
HowardWolosky, line 485: Here's another example (where line 460 won't have an invocation line prior to it)

GitHubProjects.ps1
```diff
@@ -390,6 +386,13 @@ filter New-GitHubProject
         $hashBody.add('body', $Description)
     }
 
+    if (-not $PSCmdlet.ShouldProcess($ProjectName, 'Create GitHub Project'))
+    {
+        return
+    }
+
+    Write-InvocationLog
```
HowardWolosky, line 394: And here we have two `Write-InvocationLog` calls...one here, and one still at the top of the function (line 346).

GitHubRepositories.ps1
```diff
@@ -214,6 +211,13 @@ filter New-GitHubRepository
     if ($PSBoundParameters.ContainsKey('DeleteBranchOnMerge')) { $hashBody['delete_branch_on_merge'] = $DeleteBranchOnMerge.ToBool() }
     if ($PSBoundParameters.ContainsKey('IsTemplate')) { $hashBody['is_template'] = $IsTemplate.ToBool() }
 
+    if (-not $PSCmdlet.ShouldProcess($RepositoryName, 'Create GitHub Repository'))
+    {
+        return
+    }
+
+    Write-InvocationLog
```
HowardWolosky, line 219: Another instance where we have error messages that might be logged before it.

GitHubUsers.ps1
```diff
@@ -387,6 +380,13 @@ function Set-GitHubProfile
     if ($PSBoundParameters.ContainsKey('Bio')) { $hashBody['bio'] = $Bio }
     if ($PSBoundParameters.ContainsKey('Hireable')) { $hashBody['hireable'] = $Hireable.ToBool() }
 
+    if (-not $PSCmdlet.ShouldProcess('Update Current GitHub User'))
+    {
+        return
+    }
+
+    Write-InvocationLog
```
HowardWolosky, line 388: Double `Write-InvocationLog` (here and at 372)

GitHubProjectCards.ps1
X-Guardian: But the invocation may never have happened if the user passed `whatif` to the function or selected `No` at confirm.

GitHubProjects.ps1
X-Guardian: First one removed.

GitHubUsers.ps1
X-Guardian: First one removed.

GitHubProjectCards.ps1
HowardWolosky: But the user _did_ call it.  They just happened to call it with `-WhatIf`.  It still happened, and the invocation log will capture the fact that it was called with `-WhatIf`.  Some of these functions will print out errors as a result of processing the arguments before they ever call `Invoke-GHRestMethod`.  I contend that we want to see the invocation line before any such error messages in the logs in order to have context as to why those errors are happening.

**Scenario:  User runs `Move-GitHubProjectCard -Card -Top -Bottom -WhatIf`**
**With your current change**, there will be an error reported on console and in the log with "`You must use one (and only one) of the parameters Top, Bottom, or After.`"  This will be a disembodied error, because there will be nothing in the log to indicate _what_ caused that error to happen.

**If we keep the `Write-InvocationLog` to the top**, then it will capture the call as:
```
2020-07-15 10:35:34 : Howard : VERBOSE : [0.14.0] Executing: Move-GitHubProjectCard -Card -Top -Bottom -WhatIf
2020-07-15 10:35:34 : Howard : ERROR : You must use one (and only one) of the parameters Top, Bottom or After.
```
which is much more useful.  We see the error, but we also see the calling line to understand _why_ the error happened.

Users have the ability to filter out entries in the log file that contain `-WhatIf` if they don't find those relevant, but they have no way of getting context to disembodied errors if we postpone the invocation log entry until after user confirmation, unless we move the user confirmation to the _very top_ of the function, and we've previously agreed that the parameter processing _before_ the user confirmation has value.

**I do see your point however** for scenarios where `$ConfirmPreference -ne 'None'` and a user neither passed-in `-Force` nor `-Confirm:$false`.  In those scenarios, I believe your concern is that the invocation log entry will be there without any indication on whether or not the user selected yes or no.  _That's your concern, right?_  If so, then I think the existence (or absence) of the subsequent [`Invoke-GHRestMethod` description log entry](https://github.com/microsoft/PowerShellForGitHub/blob/fcfe0c73b03be8f60ed0c75ced572240ff6769e1/GitHubCore.ps1#L206) can act as the confirmation that the action was confirmed by the user to continue.

_Aside_: You still have `Write-InvocationLog` twice in this function...once at the top and then again at 618.

HowardWolosky: /cc: @X-Guardian - not sure if you had seen this response above earlier.

X-Guardian: So is the final conclusion. to have the `Write-InvocationLog` at the top of each and every function?

HowardWolosky: That is my suggestion given everything else.  It's also how I proceeded when applying the concepts from this PR to #110 and #172.

X-Guardian: Ok, I'll bring this PR inline with this then.

X-Guardian:
@HowardWolosky, this is ready for the next review.

Telemetry.ps1
```diff
@@ -139,7 +139,8 @@ function Invoke-SendTelemetryEvent
         Invoke-* methods share a common base code.  Leaving this as-is to make this file
         easier to share out with other PowerShell projects.
 #>
-    [CmdletBinding(SupportsShouldProcess)]
```
HowardWolosky, line 142: This change is causing the tests to fail right now, because we still have the check at line 157 that calls `$PSCmdlet.ShouldProcess()`:

```
Invoke-SendTelemetryEvent' calls ShouldProcess/ShouldContinue but does not have the ShouldProcess attribute.

Stack trace
at line: 157 in /home/vsts/work/1/s/Telemetry.ps1
 ShouldProcess
 Severity: Warning
```

GitHubConfiguration.ps1
```diff
@@ -267,6 +265,11 @@ function Set-GitHubConfiguration
         }
     }
 
+    if (-not $PSCmdlet.ShouldProcess('GitHubConfiguration', 'Save'))
```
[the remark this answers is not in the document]
X-Guardian: Moved `ShouldProcess` to before line 252.

Telemetry.ps1
X-Guardian: Removed `ShouldProcessBlock`

HowardWolosky:
There are 7 UT failures with this change (all 7 are consistently failing across all 4 targets, 2 additionals are one-off issues I believe).

All 7 are in the GitHubBranches.tests.ps1, most likely around usage of `-WhatIf` in `Get-*` functions...

```
Expected an exception, with message 'Origin branch Missing-Branch not found' to be thrown, but the message was 'A parameter cannot be found that matches parameter name 'Whatif'.'. from /home/vsts/work/1/s/GitHubBranches.ps1:300 char:53
at <ScriptBlock>, /home/vsts/work/1/s/Tests/GitHubBranches.tests.ps1: line 221
at <ScriptBlock>, /home/vsts/work/1/s/Tests/GitHubBranches.tests.ps1: line 235
```

X-Guardian:
@HowardWolosky, I've remove the redundant `WhatIf/Confirm` parameters. Can you re-run the CI?

HowardWolosky:
Nice to get this in. Thanks so much for spearheading this effort, @X-Guardian!

PR #306, opened 2020-12-17
Fix labels in events being treated as full label objects

joseartrivera:
#### Description
The Get-GitHubEvent is currently broken if the issue contains and event that includes the "label" or "labels" property. This is because Add-GitHubEventAdditionalProperties treats the labels in the event object as full fledged label objects that normally contain a URL property and other additional pieces of data. The label property in events is a lighter-weight object that simply contains the string for the label and color that was changed, it is not the full object that you would typically receive from a call like Get-GithubLabel.

#### Issues Fixed
none

#### References
The docs do not explicitly call this out, but you can check the payload in fiddler for verificaiton.

#### Checklist

- [ ] [Static analysis](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#static-analysis) is reporting back clean.
- [ ] Unit tests were added/updated and are all passing. See [testing guidelines](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#testing).  This includes making sure that all pipeline input variations have been covered.

### DIFF
--- a/GitHubLabels.ps1
+++ b/GitHubLabels.ps1
@@ -1330,6 +1330,9 @@ filter Add-GitHubLabelAdditionalProperties
     .PARAMETER TypeName
         The type that should be assigned to the object.
 
+    .PARAMETER Uri
+        Repoistory URL to include as the RepositoryUrl additional properties, typically used if the type is GitHub.LabelSummary rather than GitHub.Label
+
     .INPUTS
         [PSCustomObject]
 
@@ -1347,7 +1350,10 @@ filter Add-GitHubLabelAdditionalProperties
         [PSCustomObject[]] $InputObject,
 
         [ValidateNotNullOrEmpty()]
-        [string] $TypeName = $script:GitHubLabelTypeName
+        [string] $TypeName = $script:GitHubLabelTypeName,
+
+        [Alias('RepositoryUrl')]
+        [string] $Uri
     )
 
     foreach ($item in $InputObject)
@@ -1356,7 +1362,15 @@ filter Add-GitHubLabelAdditionalProperties
 
         if (-not (Get-GitHubConfiguration -Name DisablePipelineSupport))
         {
-            $elements = Split-GitHubUri -Uri $item.url
+            if ([System.String]::IsNullOrEmpty($item.url) -and $PSBoundParameters.ContainsKey('Uri'))
+            {
+                $elements = Split-GitHubUri -Uri $Uri
+            }
+            else
+            {
+                $elements = Split-GitHubUri -Uri $item.url
+            }
+
             $repositoryUrl = Join-GitHubUri @elements
             Add-Member -InputObject $item -Name 'RepositoryUrl' -Value $repositoryUrl -MemberType NoteProperty -Force
 
--- a/GitHubLabels.ps1
+++ b/GitHubLabels.ps1
@@ -1327,12 +1327,12 @@ filter Add-GitHubLabelAdditionalProperties
     .PARAMETER InputObject
         The GitHub object to add additional properties to.
 
-    .PARAMETER TypeName
-        The type that should be assigned to the object.
-
     .PARAMETER RepositoryUrl
         Optionally supplied if the Label object doesn't have this value already
         (as is the case for GitHub.LabelSummary).
+
+    .PARAMETER TypeName
+        The type that should be assigned to the object.
 
     .INPUTS
         [PSCustomObject]
--- a/GitHubLabels.ps1
+++ b/GitHubLabels.ps1
@@ -1330,8 +1330,9 @@ filter Add-GitHubLabelAdditionalProperties
     .PARAMETER TypeName
         The type that should be assigned to the object.
 
-    .PARAMETER Uri
-        Repoistory URL to include as the RepositoryUrl additional properties, typically used if the type is GitHub.LabelSummary rather than GitHub.Label
+    .PARAMETER RepositoryUrl
+        Optionally supplied if the Label object doesn't have this value already
+        (as is the case for GitHub.LabelSummary).
 
     .INPUTS
         [PSCustomObject]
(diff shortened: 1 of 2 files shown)


HowardWolosky (changes requested):
Thanks for the report.  I was able to repro your issue by calling 

```powershell
Get-GitHubEvent -OwnerName microsoft -RepositoryName PowerShellForGitHub -Issue 306
```

What you supplied would indeed fix the specific issue you're running into, but I don't know if it's actually the _right_ fix.

I think the more complete fix would be to create a new `Type` called `GitHub.LabelSummary` for the reduce content, similar to what we do for [`GitHub.Team`/`GitHub.TeamSummary`](https://github.com/microsoft/PowerShellForGitHub/blob/d4997057f8b1234ea1aabeb4fb6742148d3afaaf/GitHubTeams.ps1#L5-L6), and then update `Add-GitHubLabelAdditionalProperties` to optionally take in the `OwnerName` and `RepositoryName`.  If supplied, and `$item.url` is not defined, then those should be used for forming the `$repositoryUrl`.  Then you'd update [`Add-GitHubEventAdditionalProperties`](https://github.com/microsoft/PowerShellForGitHub/blob/d4997057f8b1234ea1aabeb4fb6742148d3afaaf/GitHubEvents.ps1#L238-L246) to pass in the `ReposistoryUrl` and `OwnerName` along with the new `TypeName`.

That should fix the bug while also ensuring individual labels can still be passed via the pipeline into other functions.

joseartrivera:
> Thanks for the report. I was able to repro your issue by calling
> 
> ```powershell
> Get-GitHubEvent -OwnerName microsoft -RepositoryName PowerShellForGitHub -Issue 306
> ```
> 
> What you supplied would indeed fix the specific issue you're running into, but I don't know if it's actually the _right_ fix.
> 
> I think the more complete fix would be to create a new `Type` called `GitHub.LabelSummary` for the reduce content, similar to what we do for [`GitHub.Team`/`GitHub.TeamSummary`](https://github.com/microsoft/PowerShellForGitHub/blob/d4997057f8b1234ea1aabeb4fb6742148d3afaaf/GitHubTeams.ps1#L5-L6), and then update `Add-GitHubLabelAdditionalProperties` to optionally take in the `OwnerName` and `RepositoryName`. If supplied, and `$item.url` is not defined, then those should be used for forming the `$repositoryUrl`. Then you'd update [`Add-GitHubEventAdditionalProperties`](https://github.com/microsoft/PowerShellForGitHub/blob/d4997057f8b1234ea1aabeb4fb6742148d3afaaf/GitHubEvents.ps1#L238-L246) to pass in the `ReposistoryUrl` and `OwnerName` along with the new `TypeName`.
> 
> That should fix the bug while also ensuring individual labels can still be passed via the pipeline into other functions.

I thought of this approach too, but I thought it was a mistake that would add too much complexity if we try and treat this as a Label object when that was not the original intention / format of the original Github API, but I can go this route if you prefer.

joseartrivera:
> Thanks for the report. I was able to repro your issue by calling
> 
> ```powershell
> Get-GitHubEvent -OwnerName microsoft -RepositoryName PowerShellForGitHub -Issue 306
> ```
> 
> What you supplied would indeed fix the specific issue you're running into, but I don't know if it's actually the _right_ fix.
> 
> I think the more complete fix would be to create a new `Type` called `GitHub.LabelSummary` for the reduce content, similar to what we do for [`GitHub.Team`/`GitHub.TeamSummary`](https://github.com/microsoft/PowerShellForGitHub/blob/d4997057f8b1234ea1aabeb4fb6742148d3afaaf/GitHubTeams.ps1#L5-L6), and then update `Add-GitHubLabelAdditionalProperties` to optionally take in the `OwnerName` and `RepositoryName`. If supplied, and `$item.url` is not defined, then those should be used for forming the `$repositoryUrl`. Then you'd update [`Add-GitHubEventAdditionalProperties`](https://github.com/microsoft/PowerShellForGitHub/blob/d4997057f8b1234ea1aabeb4fb6742148d3afaaf/GitHubEvents.ps1#L238-L246) to pass in the `ReposistoryUrl` and `OwnerName` along with the new `TypeName`.
> 
> That should fix the bug while also ensuring individual labels can still be passed via the pipeline into other functions.

I updated the code, there is no need to update the params to GitHubEventAdditionalProperties as the repository url is already available from the base event object.

HowardWolosky (changes requested):
Thanks for the update.  Some minor additional updates being requested here.

GitHubLabels.ps1
```diff
@@ -1330,6 +1330,10 @@ filter Add-GitHubLabelAdditionalProperties
     .PARAMETER TypeName
         The type that should be assigned to the object.
 
+    .PARAMETER RepositoryUrl
+        Optionally supplied if the Label object doesn't have this value already
+        (as is the case for GitHub.LabelSummary).
```
HowardWolosky, line 1335: These should be listed in the same order as the parameters.

```suggestion
    .PARAMETER RepositoryUrl
        Optionally supplied if the Label object doesn't have this value already
        (as is the case for GitHub.LabelSummary).

    .PARAMETER TypeName
        The type that should be assigned to the object.
```

GitHubLabels.ps1
```diff
@@ -1356,8 +1362,16 @@ filter Add-GitHubLabelAdditionalProperties
...
+            if ([System.String]::IsNullOrEmpty($item.url) -and $PSBoundParameters.ContainsKey('RepositoryUrl'))
+            {
+                $repositoryUrl = $RepositoryUrl
+            }
+            else
+            {
+                $elements = Split-GitHubUri -Uri $item.url
+                $repositoryUrl = Join-GitHubUri @elements
+            }
+
```
HowardWolosky, line 1374: Simplifying

```suggestion
            if (-not [System.String]::IsNullOrEmpty($item.url))
            {
                $elements = Split-GitHubUri -Uri $item.url
                $RepositoryUrl = Join-GitHubUri @elements
            }

```

GitHubLabels.ps1
```diff
@@ -1356,8 +1362,16 @@ filter Add-GitHubLabelAdditionalProperties
...
             Add-Member -InputObject $item -Name 'RepositoryUrl' -Value $repositoryUrl -MemberType NoteProperty -Force
```
HowardWolosky, line 1375:
```suggestion
            Add-Member -InputObject $item -Name 'RepositoryUrl' -Value $RepositoryUrl -MemberType NoteProperty -Force
```

HowardWolosky (approved):
Thanks for the update.
There are some additional changes to make, but I've already supplied the fixes as suggestions and can apply them myself once this review is submitted.

HowardWolosky (approved):

PR #364, opened 2022-08-15
Support for Set-GitHubContent to Upload binary files

### LINKED ISSUE #335: Set-GitHubContent should allow upload of binary files
#### Feature Idea Summary
The current implementation of `Set-GitHubContent` only allows for text based content to be created. A simple change could allow for any arbitrary file to be uploaded.

#### Requested Assignment
I will start a PR but will need some guidance.

#### Operating System
OsName               : Microsoft Windows 10 Enterprise
OsOperatingSystemSKU : EnterpriseEdition
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsBuildLabEx    : 19041.1.amd64fre.vb_release.191206-1406
OsLanguage           : en-US
OsMuiLanguages       : {en-US}

#### PowerShell Version
PSVersion                      7.1.3
GitCommitId                    7.1.3
OS                             Microsoft Windows 10.0.19043

#### Module Version
Running: 0.16.0
Installed: 0.16.0

StanleyGoldman:
#### Description
As mentioned in #335, this pull request adds functionality to upload binary content with Set-GitHubContent
#336 got closed to due inactivity. Apologies. I'm still very interested in getting this code merged.

#### Issues Fixed
Fixes #335

#### Checklist

- [ ] [Static analysis](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#static-analysis) is reporting back clean.
- [ ] New/changed code adheres to our [coding guidelines](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#coding-guidelines).
- [ ] New/changed code continues to [support the pipeline](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#pipeline-support).
- [ ] Unit tests were added/updated and are all passing. See [testing guidelines](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#testing).  This includes making sure that all pipeline input variations have been covered.

### DIFF
--- a/GitHubContents.ps1
+++ b/GitHubContents.ps1
@@ -387,7 +387,7 @@ filter Set-GitHubContent
 
     if (-not [String]::IsNullOrWhitespace($ContentPath))
     {
-        $encodedContent = [Convert]::ToBase64String((Get-Content -Path $ContentPath -AsByteStream -Raw))
+        $encodedContent = [Convert]::ToBase64String([System.IO.File]::ReadAllBytes($ContentPath))
     }
     else
     {
--- a/GitHubContents.ps1
+++ b/GitHubContents.ps1
@@ -231,6 +231,9 @@ filter Set-GitHubContent
     .PARAMETER Content
         The new file content.
 
+    .PARAMETER ContentPath
+        The local path to file content.
+
     .PARAMETER Sha
         The SHA value of the current file if present. If this parameter is not provided, and the
         file currently exists in the specified branch of the repo, it will be read to obtain this
@@ -322,7 +325,6 @@ filter Set-GitHubContent
         [string] $CommitMessage,
 
         [Parameter(
-            Mandatory,
             Position = 4)]
         [string] $Content,
 
@@ -340,9 +342,11 @@ filter Set-GitHubContent
 
         [string] $AuthorEmail,
 
+        [string] $ContentPath,
+
         [switch] $PassThru,
 
-        [string] $AccessToken
+        [string] $AccessToken,
     )
 
     Write-InvocationLog
@@ -358,7 +362,12 @@ filter Set-GitHubContent
 
     $uriFragment = "/repos/$OwnerName/$RepositoryName/contents/$Path"
 
-    $encodedContent = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($Content))
+    if ($ContentPath) {
+        $encodedContent = [Convert]::ToBase64String((Get-Content -path $ContentPath -AsByteStream -Raw))
+    }
+    else {
+        $encodedContent = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($Content))
+    }
 
     $hashBody = @{
         message = $CommitMessage
--- a/GitHubContents.ps1
+++ b/GitHubContents.ps1
@@ -385,7 +385,7 @@ filter Set-GitHubContent
 
     $uriFragment = "/repos/$OwnerName/$RepositoryName/contents/$Path"
 
-    if ($ContentPath)
+    if (-not [String]::IsNullOrWhitespace($ContentPath))
     {
         $encodedContent = [Convert]::ToBase64String((Get-Content -Path $ContentPath -AsByteStream -Raw))
     }
--- a/GitHubContents.ps1
+++ b/GitHubContents.ps1
@@ -385,7 +385,7 @@ filter Set-GitHubContent
 
     $uriFragment = "/repos/$OwnerName/$RepositoryName/contents/$Path"
 
-    if (-not [String]::IsNullOrWhitespace($ContentPath))
+    if ($PSBoundParameters.ContainsKey('ContentPath')
     {
         $ContentPath = Resolve-UnverifiedPath -Path $ContentPath
         $encodedContent = [Convert]::ToBase64String([System.IO.File]::ReadAllBytes($ContentPath))
--- a/GitHubContents.ps1
+++ b/GitHubContents.ps1
@@ -362,10 +362,12 @@ filter Set-GitHubContent
 
     $uriFragment = "/repos/$OwnerName/$RepositoryName/contents/$Path"
 
-    if ($ContentPath) {
-        $encodedContent = [Convert]::ToBase64String((Get-Content -path $ContentPath -AsByteStream -Raw))
+    if ($ContentPath)
+    {
+        $encodedContent = [Convert]::ToBase64String((Get-Content -Path $ContentPath -AsByteStream -Raw))
     }
-    else {
+    else
+    {
         $encodedContent = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($Content))
     }
 
--- a/GitHubContents.ps1
+++ b/GitHubContents.ps1
@@ -385,7 +385,7 @@ filter Set-GitHubContent
 
     $uriFragment = "/repos/$OwnerName/$RepositoryName/contents/$Path"
 
-    if ($PSBoundParameters.ContainsKey('ContentPath')
+    if ($PSBoundParameters.ContainsKey('ContentPath'))
     {
         $ContentPath = Resolve-UnverifiedPath -Path $ContentPath
         $encodedContent = [Convert]::ToBase64String([System.IO.File]::ReadAllBytes($ContentPath))
--- a/GitHubContents.ps1
+++ b/GitHubContents.ps1
@@ -232,7 +232,7 @@ filter Set-GitHubContent
         The new file content.
 
     .PARAMETER ContentPath
-        The local path to file content.
+        The local path to a file whose content should replace what is currently stored remotely in the repo at the Path location.
 
     .PARAMETER Sha
         The SHA value of the current file if present. If this parameter is not provided, and the
--- a/GitHubContents.ps1
+++ b/GitHubContents.ps1
@@ -387,6 +387,7 @@ filter Set-GitHubContent
 
     if (-not [String]::IsNullOrWhitespace($ContentPath))
     {
+        $ContentPath = Resolve-UnverifiedPath -Path $ContentPath
         $encodedContent = [Convert]::ToBase64String([System.IO.File]::ReadAllBytes($ContentPath))
     }
     else
--- a/GitHubContents.ps1
+++ b/GitHubContents.ps1
@@ -28,7 +28,7 @@ filter Get-GitHubContent
         them individually.
 
     .PARAMETER Path
-        The file path for which to retrieve contents
+        The remote file path in the repository whose content should be changed.
 
     .PARAMETER BranchName
         The branch, or defaults to the default branch of not specified.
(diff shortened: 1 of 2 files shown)


HowardWolosky:
Thanks @StanleyGoldman -- I have this queued for review this week.

GitHubContents.ps1
```diff
@@ -231,6 +231,9 @@ filter Set-GitHubContent
     .PARAMETER Content
         The new file content.
 
+    .PARAMETER ContentPath
+        The local path to file content.
```
HowardWolosky, line 235:
```suggestion
        The local path to a file whose content should replace what is currently stored remotely in the repo at the Path location.
```

GitHubContents.ps1
```diff
@@ -358,7 +385,14 @@ filter Set-GitHubContent
 
     $uriFragment = "/repos/$OwnerName/$RepositoryName/contents/$Path"
 
-    $encodedContent = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($Content))
+    if ($ContentPath)
```
HowardWolosky, line 388: I generally don't link to do boolean checks on strings.  This should either be something like:

```suggestion
    if (-not [String]::IsNullOrWhitespace($ContentPath))
```

or

```suggestion
    if (-not [String]::IsNullOrWhitespace($ContentPath))
```

or

```suggestion
    if ($PSCmdlet.ParameterSetName -in @('ElementsContent', 'UriContent'))
```

I think the **second one** is the most clear on the expectation that we should only go into this branch _if this parameter was provided at all_.

Tests/GitHubContents.tests.ps1
```diff
@@ -335,6 +335,67 @@ try
...
+                $authorEmail = 'jane.doe@testdomain.com'
+
+                $contentPath = New-TemporaryFile
+                $content | Out-File -FilePath $contentPath
+
+                $setGitHubContentParms = @{
+                    Path = "$filePath/$fileName"
+                    CommitMessage = $commitMessage
+                    Branch = $branchName
+                    ContentPath = $contentPath
```
HowardWolosky, line 357: Technically speaking, `$contentPath` at this point in time is a `[FileInfo]`, not a `[String]`.  So, what you want here is:

```suggestion
                    ContentPath = $contentPath.FullName
```

Tests/GitHubContents.tests.ps1
```diff
@@ -335,6 +335,67 @@ try
...
+
+            It 'Should have written the correct content' {
+                $getGitHubContentParms = @{
+                    Path = "$filePath/$fileName"
+                    Uri = $repo.svn_url
+                    MediaType = 'Raw'
+                    ResultAsString = $true
+                }
+
+                $writtenContent = Get-GitHubContent @getGitHubContentParms
```
HowardWolosky, line 389: Hmmmm....I _think_ that running another command while within the `It` block goes against the Pester 5 guidelines.   @tigerfansga, can you chime in here on this?  I want to make sure that we're not doing something that is going to break all your hard work in converting the tests over.

HowardWolosky (changes requested):
Getting close -- thanks for reviving this change.  Looking forward to getting it merged in.

Tests/GitHubContents.tests.ps1
tigerfansga: > Hmmmm....I _think_ that running another command while within the `It` block goes against the Pester 5 guidelines. @tigerfansga, can you chime in here on this? I want to make sure that we're not doing something that is going to break all your hard work in converting the tests over.

@HowardWolosky this should be fine. Running commands in an It block is fine. The item to be careful is Pester v5 creates a new scope before calling an It block. So any variables defined in the It block are limited to that It block. In this case, he is only using the variable inside a single It block. See the *Run* section of [Discovery and Run](https://pester.dev/docs/usage/discovery-and-run)

This code could be consolidated to 
```powershell
$content | Should -Be (Get-GitHubContent @getGitHubContentParms)
```
If any test changes are merged before my changes are merged, I will likely need to update my branch and validate the changes work as merged.

StanleyGoldman:
Heya, I'd still love to get this change in.

HowardWolosky:
> Heya, I'd still love to get this change in.

That's great news.  I'm re-opening the PR.  Your next step is to address the PR feedback.

Tests/GitHubContents.tests.ps1
StanleyGoldman: Let me know if anything needs to change here.

StanleyGoldman:
@HowardWolosky ready when you are.

Tests/GitHubContents.tests.ps1
```diff
@@ -226,6 +226,121 @@ Describe 'Getting file and folder content' {
...
+
+                $result = Set-GitHubContent @setGitHubContentParms -PassThru
+            }
+
+            It 'Should have the expected type and additional properties' {
+                $result.PSObject.TypeNames[0] | Should -Be 'GitHub.Content'
+                $result.content.name | Should -Be $fileName
+                $result.content.path | Should -Be "$filePath/$fileName"
+                $result.content.url | Should -Be ("https://api.github.com/repos/$($script:ownerName)" +
+                    "/$repoName/contents/$filePath/$($fileName)?ref=$BranchName")
```
HowardWolosky, line 264: `$repoName` doesn't exist in this context.  In short, you put these tests in the wrote `Describe` block. They should be part of the block starting at 393 (`Describe 'GitHubContents/Set-GitHubContent' {`)

Tests/GitHubContents.tests.ps1
```diff
@@ -226,6 +226,121 @@ Describe 'Getting file and folder content' {
...
```
HowardWolosky, line 322:
```suggestion
                    "/$repoName/contents/$filePath/$($fileName)?ref=$branchName")
```

Tests/GitHubContents.tests.ps1
StanleyGoldman: Block moved.

StanleyGoldman:
Changes made.

As GitHub staff I was a bit reluctant to run integration tests that used GitHub as a testbed. I was hoping there might be some GitHub Action (my team, shameless plug) that kicked in when you started to look at my PR that would catch my mistakes. Alas, you caught me. Apologies 🙏.

GitHubContents.ps1
```diff
@@ -358,7 +385,14 @@ filter Set-GitHubContent
 
     $uriFragment = "/repos/$OwnerName/$RepositoryName/contents/$Path"
 
-    $encodedContent = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($Content))
+    if (-not [String]::IsNullOrWhitespace($ContentPath))
+    {
+        $encodedContent = [Convert]::ToBase64String([System.IO.File]::ReadAllBytes($ContentPath))
```
HowardWolosky, line 390: Now that you aren't accessing the file directly through a PowerShell function, you need to proactively resolve this filepath, certain filepaths won't be processed correctly when passed-in to a .NET function like this unless we ensure it's full resolved.

I'd suggest you do this first:

```suggestion
        $ContentPath = Resolve-UnverifiedPath -Path $ContentPath
        $encodedContent = [Convert]::ToBase64String([System.IO.File]::ReadAllBytes($ContentPath))
```

GitHubContents.ps1
```diff
@@ -358,7 +385,14 @@ filter Set-GitHubContent
 
     $uriFragment = "/repos/$OwnerName/$RepositoryName/contents/$Path"
 
-    $encodedContent = [Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($Content))
+    if (-not [String]::IsNullOrWhitespace($ContentPath))
```
HowardWolosky, line 388: Given that what you really mean here is "if our parameter set included `ContentPath`, let's go in here", it would be more clear (and correct) if you did this.

```suggestion
    if ($PSBoundParameters.ContainsKey('ContentPath')
```

Otherwise, what you're saying is, if `$ContentPath` only has whitespace, let's use `$Content` instead, but we won't have that at all because that wasn't in our parameter set.

Tests/GitHubContents.tests.ps1
```diff
@@ -399,6 +399,119 @@ Describe 'GitHubContents/Set-GitHubContent' {
...
+
+            $content | Should -Be $writtenContent
+        }
+
+        AfterAll {
+            Remove-Item -Path $contentPath
+        }
+    }
+
+    Context 'When overwriting file content by path' {
```
HowardWolosky, line 463: I don't think this test is actually testing this.  This is essentially a duplicate of the previous test, because you had an `AfterAll` in the previous `Context` that removed the file. You'll probably want to refactor the tests to be within the same `Context` so that you can do what you're trying to do.

HowardWolosky (changes requested):
Small bit of a additional feedback.  Thanks for the effort here, and apologies for delay in review.

GitHubContents.ps1
StanleyGoldman: That makes sense

GitHubContents.ps1
StanleyGoldman: `$PSBoundParameters` Thank you, you are teaching me some PowerShell now.

Tests/GitHubContents.tests.ps1
StanleyGoldman: I meant this text context to be an analogue to 'When overwriting file content', but I had some bad copy-pasta going on. I believe I have fixed the issue.

Tests/GitHubContents.tests.ps1
```diff
@@ -399,6 +399,132 @@ Describe 'GitHubContents/Set-GitHubContent' {
...
+            $branchName = 'master'
+            $committerName = 'John Doe'
+            $committerEmail = 'john.doe@testdomain.com'
+            $authorName = 'Jane Doe'
+            $authorEmail = 'jane.doe@testdomain.com'
+
+            . (Join-Path -Path $moduleRootPath -ChildPath 'Helpers.ps1')
+
+            $contentPath = New-TemporaryFile
+            $contentPath = Resolve-UnverifiedPath -Path $contentPath
```
HowardWolosky, line 417: It's not clear to me why you're using `Resolve-UnverifiedPath` here (which is requiring you to bring in Helpers.ps1).  `New-TemporaryFile` gives back a FileInfo which provides you the full path to the newly created file.  You should simply be able to reference `$contentPath.FullName` in `[System.IO.File]::WriteAllText` and be good to go.  `Resolve-UnverifiedPath` is needed when there's the possibility we'll be working with a partial/relative path.

HowardWolosky (changes requested):
Apologies for taking so long to finally review this update.
The product change looks good.
On the test side, the first test is also good.
The only concern that I have is that it looks like you intended for your second test to be dependent on the first test (in that you're hoping to validate that you can overwrite existing content that was written by the previous test.  The issue of that is that each of these Contexts should be able to run independently.  If you want a context that is testing overwriting existing content, then you should have that happen within the scope of a single Context.

Ideally we should be able to repro a test bug by running just the failing test, which wouldn't be possible if the failure only repro'd if something needed to happen in a separate context first.

PR #381, opened 2022-12-15
Revert "New/Set-GitHubRepository: Add AllowAutoMerge and UseSquashPrTitleAsDefault Parameters"

HowardWolosky:
Reverts microsoft/PowerShellForGitHub#358

It turns out that this isn't quite working right.

The `use_squash_pr_title_as_default` parameter [has been deprecated per the documentation](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#update-a-repository) in favor of `squash_merge_commit_title ` which works a bit differently.

Additionally, there appears to be a GitHub-side bug at the moment with changing the value of `allow_auto_merge` (it does not honor the request being sent).

Given this, for the time being we'll revert the change until we know that the feature is working correctly.

/cc: @X-Guardian

### DIFF
--- a/GitHubRepositories.ps1
+++ b/GitHubRepositories.ps1
@@ -83,13 +83,6 @@ filter New-GitHubRepository
         By default, rebase-merge pull requests will be allowed.
         Specify this to disallow.
 
-    .PARAMETER AllowAutoMerge
-        Specifies whether to allow auto-merge on pull requests.
-
-    .PARAMETER UseSquashPrTitleAsDefault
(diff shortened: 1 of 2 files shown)


X-Guardian:
I will raise new PRs for these, splitting each one into its own PR. I have just tested my original code for the `allow_auto_merge` setting on both the `new` and `set` cmdlets and it is working fine for me now, so it was probably a transient issue with GitHub when you tested it.

Issue #406, opened 2023-04-28
API Coverage: Add Support for Codespaces


vercellone:
#### Feature Idea Summary
API Coverage #70
Add support for [Codespaces](https://docs.github.com/en/rest/codespaces).

#### Feature Idea Additional Details
I'll be targeting coverage for all of the endpoints documented in the following areas:
- https://docs.github.com/en/rest/codespaces/codespaces
- https://docs.github.com/en/rest/codespaces/organizations

#### Requested Assignment
I'm confident I can deliver some (if not all) of these, because I recently wrote pwsh 7 functions covering them all.  However, I will likely have questions along the way.

I've started with Get-GitHubCodespace in my [feature/GitHubCodespaces](https://github.com/microsoft/PowerShellForGitHub/compare/master...vercellone:PowerShellForGitHub:feature/GitHubCodespaces) branch.

#### Operating System
OsName               : Microsoft Windows 11 Pro
OsOperatingSystemSKU : 48
OsArchitecture       : 64-bit
WindowsVersion       : 2009
WindowsBuildLabEx    : 22621.1.amd64fre.ni_release.220506-1250
OsLanguage           : en-US
OsMuiLanguages       : {en-US}

#### PowerShell Version
Name                           Value
----                           -----
PSVersion                      7.3.4
PSEdition                      Core
GitCommitId                    7.3.4
OS                             Microsoft Windows 10.0.22621
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

#### Module Version
Running: 0.16.1
Installed: 0.16.1

vercellone:
I committed 3 functions to my [feature/GitHubCodespaces](https://github.com/microsoft/PowerShellForGitHub/compare/master...vercellone:PowerShellForGitHub:feature/GitHubCodespaces) branch.
- Start-GitHubCodespace
- Stop-GitHubCodespace
- Wait-GitHubCodespaceAction
  - This is currently a private (not exported) function that facilitates a -Wait switch for start & stop.
  - Let me know if it makes sense to export it.

Issue #70, opened 2018-12-11
Remaining work for API Completeness


HowardWolosky:
Using a single issue to track the remaining work, so as not to litter the Issues and give the wrong impression on the current state of the project, bug wise.

If a contributor is interested in making progress towards any of these items, they should just create a new Issue to track that specific work, mention this Issue, and then resolve their PR against the Issue that they created.  At that point, I'll go ahead and update this list and check off that item.

[ ] Activity
   [ ] [Core][Activity](https://developer.github.com/v3/activity/)
   [ ] [Events](https://developer.github.com/v3/activity/events/) (Issue Events are already supported)
   [ ] [Feeds](https://developer.github.com/v3/activity/feeds/)
   [ ] [Notifications](https://developer.github.com/v3/activity/notifications/)
   [ ] [Starring](https://developer.github.com/v3/activity/starring/)
   [ ] [Watching](https://developer.github.com/v3/activity/watching/)
[ ] Gists
   [ ] [Core](https://developer.github.com/v3/gists/)
   [ ] [Comments](https://developer.github.com/v3/gists/comments/)
[ ] Git Data
   [ ] [Blobs](https://developer.github.com/v3/git/blobs/)
   [ ] [Commits](https://developer.github.com/v3/git/commits/)
   [ ] [Refs](https://developer.github.com/v3/git/refs/)
   [ ] [Tags](https://developer.github.com/v3/git/tags/)
   [ ] [Trees](https://developer.github.com/v3/git/trees/)
[ ] Migrations
   [ ] [Orgs](https://developer.github.com/v3/migrations/orgs/)
   [ ] [Source Imports](https://developer.github.com/v3/migrations/source_imports/)
   [ ] [Users](https://developer.github.com/v3/migrations/users/)
[ ] Organizations
   [ ] [Core](https://developer.github.com/v3/orgs/) (Some of this work already exists)
   [ ] [Blocking Users](https://developer.github.com/v3/orgs/blocking/)
   [ ] [Blocking Users](https://developer.github.com/v3/orgs/blocking/)
   [ ] [Members](https://developer.github.com/v3/orgs/members/)
   [ ] [Outside Collaborators](https://developer.github.com/v3/orgs/outside_collaborators/)
   [ ] [Webhooks](https://developer.github.com/v3/orgs/hooks/)
[ ] Projects
   [ ] [Core](https://developer.github.com/v3/projects/)
[...]

PR #407, opened 2023-04-29
Add support for codespaces

vercellone:
#### Description
Add support for [Codespaces](https://docs.github.com/en/rest/codespaces).

#### Issues Fixed
- #70
- #406 

#### References
Endpoints ~~excluded~~ below will be added in subsequent PRs

**[Codespaces/codespaces](https://docs.github.com/en/rest/codespaces/codespaces)**
- [x] [List codespaces in a repository for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#list-codespaces-in-a-repository-for-the-authenticated-user)
- [x] [Create a codespace in a repository](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#create-a-codespace-in-a-repository)
~~[List devcontainer configurations in a repository for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#list-devcontainer-configurations-in-a-repository-for-the-authenticated-user)~~
~~[Get default attributes for a codespace](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#get-default-attributes-for-a-codespace)~~
- [x] [Create a codespace from a pull request](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#create-a-codespace-from-a-pull-request)
- [x] [List codespaces for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#list-codespaces-for-the-authenticated-user)
- [x] [Create a codespace for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#create-a-codespace-for-the-authenticated-user)
- [x] [Get a codespace for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#get-a-codespace-for-the-authenticated-user)
~~[Update a codespace for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#update-a-codespace-for-the-authenticated-user)~~
- [x] [Delete a codespace for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#delete-a-codespace-for-the-authenticated-user)
~~[Export a codespace for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#export-a-codespace-for-the-authenticated-user)~~
~~[Get details about a codespace export](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#get-details-about-a-codespace-export)~~
~~[Create a repository from an unpublished codespace](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#create-a-repository-from-an-unpublished-codespace)~~
- [x] [Start a codespace for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#start-a-codespace-for-the-authenticated-user)
- [x] [Stop a codespace for the authenticated user](https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#stop-a-codespace-for-the-authenticated-user)

**[Codespaces/organizations](https://docs.github.com/en/rest/codespaces/organizations)**
- [x] [List codespaces for the organization](https://docs.github.com/en/rest/codespaces/organizations?apiVersion=2022-11-28#list-codespaces-for-the-organization)
~~[Manage access control for organization codespaces](https://docs.github.com/en/rest/codespaces/organizations?apiVersion=2022-11-28#manage-access-control-for-organization-codespaces)~~
~~[Add users to Codespaces billing for an organization](https://docs.github.com/en/rest/codespaces/organizations?apiVersion=2022-11-28#add-users-to-codespaces-billing-for-an-organization)~~
~~[Removes users from Codespaces billing for an organization](https://docs.github.com/en/rest/codespaces/organizations?apiVersion=2022-11-28#removes-users-from-codespaces-billing-for-an-organization)~~
- [x] [List codespaces for a user in organization](https://docs.github.com/en/rest/codespaces/organizations?apiVersion=2022-11-28#list-codespaces-for-a-user-in-organization)
- [x] [Delete a codespace from the organization](https://docs.github.com/en/rest/codespaces/organizations?apiVersion=2022-11-28#delete-a-codespace-from-the-organization)
~~[Stop a codespace for an organization user](https://docs.github.com/en/rest/codespaces/organizations?apiVersion=2022-11-28#stop-a-codespace-for-an-organization-user)~~
#### Checklist

- [x] [Formatters were created](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#formatters) for any new types being added.
- [x] Unit tests were added/updated and are all passing. See [testing guidelines](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#testing).  This includes making sure that all pipeline input variations have been covered.
- [x] If desired, ensure your name is added to our [Contributors list](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#contributors)


GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+            else {
+                $telemetryProperties['UserName'] = Get-PiiSafeString -PlainText $UserName
+                $uriFragment = "orgs/$OrganizationName/members/$UserName/codespaces"
+                $description = "Getting codespaces for $OrganizationName"
+            }
+
+            break
+        }
+
+        'Repository' {
```
HowardWolosky, line 167: Your `Repository` and `Uri` cases are identical, which leads to code duplication.
Please collapse those into a single case statement. You can write it like this:

```powershell
         { $_ -in ('Repository', 'Uri') } {
                 # ...
         }
```

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+        'UriFragment' = $uriFragment
+        'Description' = $description
+        'AcceptHeader' = "$script:nebulaAcceptHeader,$script:baptisteAcceptHeader,$script:mercyAcceptHeader"
+        'AccessToken' = $AccessToken
+        'TelemetryEventName' = $MyInvocation.MyCommand.Name
+        'TelemetryProperties' = $telemetryProperties
+    }
+
+    $result = Invoke-GHRestMethodMultipleResult @params
+    if ($result.codespaces) {
```
HowardWolosky, line 206: nit: Again, open brackets (`{`) on a separate line please.  Assume I say that everywhere else in this file as well.

HowardWolosky, line 206: I prefer not using implicit `null` checks for a boolean check.  Please use `$null -ne $result.codespaces`

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+}
+
+function Start-GitHubCodespace
+{
+<#
+    .SYNOPSIS
+        Start a user's codespace.
+
+    .DESCRIPTION
+        You must authenticate using an access token with the codespace scope to use this endpoint.
```
HowardWolosky, line 220: These first two lines in the description are great entries for a `.NOTES` section at the end of this documentation section.  For now, please just duplicate your `.SYNOPSIS` here, unless you have additional details on what this is doing (as opposed to requirements for using it).

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSReviewUnusedParameter", "", Justification="PassThru is accessed indirectly via Resolve-ParameterWithDefaultConfigurationValue")]
+    param(
+        [Parameter(Mandatory,
+            ValueFromPipeline,
+            ValueFromPipelineByPropertyName)]
+        [string] $CodespaceName,
+
+        [switch] $PassThru,
+
+        [switch] $Wait,
```
HowardWolosky, line 266: The preference is that "standard" switches like `PassThru` or `Force` (along with `AccessToken`) are the last parameters in a function call.  Please fix the order here and in the docs above.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+
+        [string] $AccessToken
+    )
+
+    process
```
HowardWolosky, line 271: Why write this as a `function` with a `process` instead of just making this a `filter`?

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    return ($result | Add-GitHubCodespaceAdditionalProperties)
+}
+
+function Start-GitHubCodespace
+{
+<#
+    .SYNOPSIS
+        Start a user's codespace.
```
HowardWolosky, line 217:
```suggestion
        Start a Codespace for the currently authenticated user.
```

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+            return $result
+        }
+    }
+}
+
+function Stop-GitHubCodespace
+{
+<#
+    .SYNOPSIS
+        Stop a user's codespace.
```
HowardWolosky, line 317:
```suggestion
        Stop a Codespace for the currently authenticated user.
```

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+
+    .DESCRIPTION
+        You must authenticate using an access token with the codespace scope to use this endpoint.
```
HowardWolosky, line 320: Same feedback as for `Start-*`.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSReviewUnusedParameter", "", Justification="PassThru is accessed indirectly via Resolve-ParameterWithDefaultConfigurationValue")]
+    param(
+        [Parameter(Mandatory,
+            ValueFromPipeline,
+            ValueFromPipelineByPropertyName)]
+        [string] $CodespaceName,
+
+        [switch] $PassThru,
+
+        [switch] $Wait,
```
HowardWolosky, line 366: Same feedback here.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+
+        [string] $AccessToken
+    )
+
+    process
```
HowardWolosky, line 371: Same feedback

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+        $telemetryProperties = @{
+            'CodespaceName' = Get-PiiSafeString -PlainText $CodespaceName
+        }
+
+        $params = @{
+            'UriFragment' = "user/codespaces/$CodespaceName/stop"
+            'Method' = 'POST'
+            'Description' = "Stop Codespace $CodespaceName"
+            'AccessToken' = $AccessToken
+            'AcceptHeader' = $script:symmetraAcceptHeader
```
HowardWolosky, line 384: This header does not seem valid for Codespaces.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+        $telemetryProperties = @{
+            'CodespaceName' = Get-PiiSafeString -PlainText $CodespaceName
+        }
+
+        $params = @{
+            'UriFragment' = "user/codespaces/$CodespaceName/start"
+            'Method' = 'POST'
+            'Description' = "Start Codespace $CodespaceName"
+            'AccessToken' = $AccessToken
+            'AcceptHeader' = $script:symmetraAcceptHeader
```
HowardWolosky, line 284: This header does not seem valid for Codespaces.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+            $description = "Getting $OwnerName/$RepositoryName/codespaces"
+
+            break
+        }
+    }
+
+    $params = @{
+        'UriFragment' = $uriFragment
+        'Description' = $description
+        'AcceptHeader' = "$script:nebulaAcceptHeader,$script:baptisteAcceptHeader,$script:mercyAcceptHeader"
```
HowardWolosky, line 199: I don't think we need these headers...none of them seem appropriate for this API.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+        [Parameter(Mandatory,
+            ValueFromPipeline,
+            ValueFromPipelineByPropertyName)]
+        [string] $CodespaceName,
+
+        [string] $AccessToken
+    )
+
+    begin {
+        # 2s minimum
```
HowardWolosky, line 447: You say 2s minimum, but you're allowing for 1 second.  Which is it?

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+
+    process
+    {
+        Write-InvocationLog
+
+        do
+        {
+            Start-Sleep -Seconds $sleepSeconds
+            $codespace = (Get-GitHubCodespace @PSBoundParameters)
+        } until($codespace.state -notmatch 'Queued|ing')
```
HowardWolosky, line 463: Can you move `until` to the next line?

HowardWolosky, line 463: This is a bit tricky.  I understand what you're doing, but can you document in a comment what the possible '*ing' states might be?

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+            Mandatory,
+            ValueFromPipeline)]
+        [AllowNull()]
+        [AllowEmptyCollection()]
+        [PSCustomObject[]] $InputObject,
+
+        [ValidateNotNullOrEmpty()]
+        [string] $TypeName = $script:GitHubCodespaceTypeName
+    )
+    foreach ($item in $InputObject)
```
HowardWolosky, line 508: nit:

```suggestion

    foreach ($item in $InputObject)
```

GitHubCodespaces.ps1
HowardWolosky, line 1: Can you please add some tests for this file as well?
They'll be limited, of course at the moment since we only have Get, Start and Stop with no way to create, but for the moment, we can at least verify that Get is returning 0 Codespaces for all the possible parameter sets, and that Start and Stop error when called.  So, it'll be a small test file, but we should still have some (albeit limited) coverage for these newly added functions.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+        }
+
+        if (Resolve-ParameterWithDefaultConfigurationValue -Name PassThru -ConfigValueName DefaultPassThru)
+        {
+            return $result
+        }
+    }
+}
+
+function Wait-GitHubCodespaceAction
```
HowardWolosky, line 413: I'm in agreement with you that this can remain an internal helper for the time being.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+
+    .INPUTS
+        GitHub.Codespace
+
+    .OUTPUTS
+        GitHub.Codespace
+
+    .EXAMPLE
+        Wait-GitHubCodespace -Name vercellone-effective-goggles-qrv997q6j9929jx8
+#>
```
HowardWolosky, line 434: Can you add something like this to the doc too please?

https://github.com/microsoft/PowerShellForGitHub/blob/43d5392642cce2ec3d45fbbb1066f7b2efc3b8c4/GitHubGists.ps1#L313-L315

HowardWolosky (changes requested):
Thanks so much for this work, @vercellone!
This was in really great shape for a first contribution.  I do have some feedback for you to address (please), but I don't anticipate it being a significant amount of work.

I'll admit that I had no idea that this GitHub feature even existed.  I'm going to have to look to see what might be involved in getting a standard Codespace set up for this project (and some of my others).

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+            $telemetryProperties['OrganizationName'] = Get-PiiSafeString -PlainText $OrganizationName
+            if ([string]::IsNullOrWhiteSpace($UserName)) {
+                $uriFragment = "orgs/$OrganizationName/codespaces"
+                $description = "Getting codespaces for $OrganizationName"
+            }
+            else {
+                $telemetryProperties['UserName'] = Get-PiiSafeString -PlainText $UserName
+                $uriFragment = "orgs/$OrganizationName/members/$UserName/codespaces"
+                $description = "Getting codespaces for $OrganizationName"
+            }
```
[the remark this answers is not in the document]
vercellone: Force of habit.  Been using hanging brackets elsewhere.  I'll Format Document.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+        them individually.
+
+    .PARAMETER OrganizationName
+        The name of the organization to retrieve the codespaces for.
+
+    .PARAMETER AccessToken
+        If provided, this will be used as the AccessToken for authentication with the
+        REST Api.  Otherwise, will attempt to use the configured value or will run unauthenticated.
+
+    .INPUTS
```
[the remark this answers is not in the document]
vercellone: Indeed.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+        [string] $RepositoryName,
+
+        [Parameter(
+            Mandatory,
+            ValueFromPipelineByPropertyName,
+            ParameterSetName='CodespaceName')]
+        [string] $CodespaceName,
+
+        [Parameter(
+            ValueFromPipelineByPropertyName,
```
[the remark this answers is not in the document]
vercellone: Good catch, thanks!

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,538 @@
...
+        Gets all of the codespaces in the PowerShell organization.
+#>
+    [CmdletBinding(DefaultParameterSetName = 'AuthenticatedUser')]
+    [OutputType({$script:GitHubCodespaceTypeName})]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSReviewUnusedParameter", "", Justification="The Uri parameter is only referenced by Resolve-RepositoryElements which get access to it from the stack via Get-Variable -Scope 1.")]
+    param(
+        [Parameter(
+            Mandatory,
+            ValueFromPipelineByPropertyName,
+            ParameterSetName='Repository')]
```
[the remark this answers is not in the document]
vercellone: np

GitHubCodespaces.ps1
vercellone: Right.  defaultAcceptHeader is good.

GitHubCodespaces.ps1
vercellone: 2s min.

GitHubCodespaces.ps1
vercellone: My understanding is that filter is simply syntactic sugar.  I'll change it.  But, I'm curious: Is  there a performance impact?  If not, seems better in terms of long-term maintenance as a function when/if you want a begin block.

GitHubCodespaces.ps1
vercellone: Tests are forthcoming.  I have a few local tests that are still evolving.  But, as you said, options are limited given the lack of New-GitHubCodespace.  I think it will be more efficient if I remedy that first.

vercellone:
**Added**
- New-GitHubCodespace
- Remove-GitHubCodespace

Keep the feedback coming.  I'm particularly interested in your take on INPUTS, -Uri, -RepositoryId, and -PullRequest usage for New.

I will work on Tests next when able.

GitHubCodespaces.ps1
vercellone: This is subject to change when support is added for [/orgs/{org}/members/{username}/codespaces/{codespace_name}/stop](https://docs.github.com/en/rest/codespaces/organizations?apiVersion=2022-11-28#stop-a-codespace-for-an-organization-user).

There is no equivalent for start.

vercellone:
**Good news**
Repositories do not require any prep to create a codespace with New-GitHubCodespace.  Of my repos, only [vercellone/my-first-codespace](https://github.com/vercellone/my-first-codespace) includes .devcontainer/ config, and I can create viable codepaces for any of my repos.  Given that, testing looks pretty straightforward.

**Bad news**
The url returned is always of the form https://api.github.com/user/codespaces/{codespace_name}, with no owner nor repo elements.  Therefore, URI + Resolve-RepositoryElements isn't viable.  I can figure something out, but suggestions are appreciated.

HowardWolosky:
> **Good news** Repositories do not require any prep to create a codespace with New-GitHubCodespace. Of my repos, only [vercellone/my-first-codespace](https://github.com/vercellone/my-first-codespace) includes .devcontainer/ config, and I can create viable codepaces for any of my repos. Given that, testing looks pretty straightforward.

Great!

 
> **Bad news** The url returned is always of the form [https://api.github.com/user/codespaces/{codespace_name}](https://api.github.com/user/codespaces/%7Bcodespace_name%7D), with no owner nor repo elements. Therefore, URI + Resolve-RepositoryElements isn't viable. I can figure something out, but suggestions are appreciated.

At what point are you encountering this issue?  In `Add-GitHubCodespacesAdditionalProperties`?  If so, you have two options there:

1. You can model something similar to what we do elsewhere where we think we may not have the necessary information in the response object -- just pass in the information we already know as an optional parameter.  We do that for Repositories here:

https://github.com/microsoft/PowerShellForGitHub/blob/6f94a9b0a37ee466c2fb457299c78de1bd371f95/GitHubRepositories.ps1#L3673-L3702

2. The other option is to leverage the fact that a `repository` object is part of the response object and you're already sending it into `Add-GitHubRepositoryAdditionalProperties`, so you can just grab `RepositoryUrl` from the modified `repository` object afterwards and then add it to the root of your new object.

If you're encontering this issue elsewhere, then I'll need more context to know what and where the problem is.

vercellone:
> **Bad news** The url returned is always of the form [https://api.github.com/user/codespaces/{codespace_name}]

I went with [Option 2](https://github.com/microsoft/PowerShellForGitHub/pull/407/commits/c2d9604bd4f9ac8dd33692285bb078abd7043105)

vercellone:
I added some tests for Get-GitHubCodespace.  I'm guessing whatever feedback you'll have will likely be applicable to the remaining tests as well.  But, I will work on them in the meantime whether I hear back or not.

vercellone:
Tests added
- Get-GitHubCodespace
  - When specifiying the Uri parameter
  - When specifiying the Uri parameter from the pipeline
- Remove-GitHubCodespace
  - When deleting a codespace for the authenticated user
- Start-GitHubCodespace
  - When starting a codespace for the authenticated user
- Stop-GitHubCodespace
  - When stopping a codespace for the authenticated user

New-GitHubCodespace tests are still forthcoming.  As are Remove-GitHubCodespace and Stop-GitHubCodespace tests for Organization parameter sets.

A -Force switch was added to Remove-GitHubCodespace, but the current ConfirmImpact is the default of Medium.  Do you think it should be set to High, or should the -Force switch be dropped?  The codespace is easily rebuilt but does maintain the state of work in progress.  I'm leaning toward medium with no Force, but could be easily swayed.

HowardWolosky:
I really appreciate all the effort you're putting into this support.  Thanks so much!

> A -Force switch was added to Remove-GitHubCodespace, but the current ConfirmImpact is the default of Medium. Do you think it should be set to High, or should the -Force switch be dropped? The codespace is easily rebuilt but does maintain the state of work in progress. I'm leaning toward medium with no Force, but could be easily swayed.

I feel like it would be best if the module was consistent with how it handles all `Remove-*` methods.  A consistent interface helps users feel comfortable and safe when exploring other parts of a module.  In this case, it should keep the `-Force` switch, and have the same behavior that the other `Remove-*` methods have when it comes to checking `-Force` and `$Confirm`, as well as `$PSCmdlet.ShouldProcess`.  Hope that helps.

BTW: I won't be looking at the updated PR again until you explicitly request it.

vercellone:
> I feel like it would be best if the module was consistent with how it handles all `Remove-*` methods.

Agreed.  ConfirmImpact set to High.

> BTW: I won't be looking at the updated PR again until you explicitly request it.

In that case I'll contribute functions for the remaining endpoints in a separate PR.  I've ~~excluded~~ them above, accordingly.  Once I get the outstanding tests and usage committed I'll let you know.

vercellone:
Tests pushed.  I still owe you Usage, but please review if able.

vercellone:
Usage committed.  Please review @HowardWolosky.

HowardWolosky:
> Usage committed. Please review @HowardWolosky.

Noted, thanks!  Will take a look this weekend.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,450 @@
...
+            }
+
+            It 'Should return an object of the correct type' {
+                $codespace | Should -BeOfType PSCustomObject
+            }
+
+            It 'Should return the correct properties' {
+                $codespace.display_name | Should -Not -BeNullOrEmpty
+                $codespace.repository.name | Should -Be $repo.name
+                $codespace.idle_timeout_minutes | Should -Be 30
```
HowardWolosky, line 221: I didn't see `30` specified anywhere in the docs as the default.  Not sure it makes sense to explicitly test for this value when the default isn't defined and you didn't specify an expected value.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,450 @@
...
+                if ($codespace)
+                {
+                    Remove-GitHubCodespace -CodespaceName $codespace.name -Confirm:$false
+                }
+            }
+        }
+
+        Context -Name 'When creating a codespace with default settings with Ref' {
+            BeforeAll {
+                $repoWithPR = Get-GitHubRepository -OrganizationName 'super-fake-not-real-org' |
```
HowardWolosky, line 236: Wait -- why would this test work at all given that this org doesn't exist?

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,450 @@
...
+                if ($codespace)
+                {
+                    Remove-GitHubCodespace -CodespaceName $codespace.name -Confirm:$false
+                }
+            }
+        }
+
+        Context -Name 'When creating a codespace with default settings from a PullRequest' {
+            BeforeAll {
+                  $repoWithPR = Get-GitHubRepository -OrganizationName 'super-fake-not-real-org' |
```
HowardWolosky, line 271: Same question here.  Why would this test work at all given that this org doesn't exist?

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,450 @@
...
+            }
+
+            It 'Should return an object of the correct type' {
+                $codespace | Should -BeOfType PSCustomObject
+            }
+
+            It 'Should return the correct properties' {
+                # $codespace.devcontainer_path | Should -Be
+                $codespace.display_name | Should -Be $newGitHubCodespaceParms.DisplayName
+                $codespace.idle_timeout_minutes | Should -Be 5
```
HowardWolosky, line 327: Here and elsewhere, I'd suggest that you reference the value from `$newGitHubCodespaceParams` to make this robust to changes in the future and avoid magic numbers.  So, instead of saying `5`, say:

```suggestion
                $codespace.idle_timeout_minutes | Should -Be $newGitHubCodespaceParams.Timeout
```

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,450 @@
...
+                $codespace.location | Should -Be $newGitHubCodespaceParms.Location
+                $codespace.machine.name | Should -Be $newGitHubCodespaceParms.Machine
+                $codespace.owner.UserName | Should -Be $script:OwnerName
+                $codespace.repository.name | Should -Be $repo.name
+                $codespace.retention_period_minutes | Should -Be 10
```
HowardWolosky, line 332:
```suggestion
                $codespace.retention_period_minutes | Should -Be $newGitHubCodespaceParams.RetentionPeriod
```

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,450 @@
...
+            }
+
+            It 'Should return an object of the correct type' {
+                $codespace | Should -BeOfType PSCustomObject
+            }
+
+            It 'Should return the correct properties' {
+                $codespace.display_name | Should -Not -BeNullOrEmpty
+                $codespace.repository.name | Should -Be $repo.name
+                $codespace.idle_timeout_minutes | Should -Be 30
```
HowardWolosky, line 360: Same comment here about not testing this value since the default isn't documented.  Here and elsewhere in the file where I may have missed adding the comment as well.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,945 @@
...
+    .PARAMETER Machine
+        Machine type to use for this codespace.
+
+    .PARAMETER NoMultipleRepoPermissions
+        Whether to authorize requested permissions to other repos from devcontainer.json.
+
+    .PARAMETER RetentionPeriod
+        Duration in minutes (up to 30 days) after codespace has gone idle in which it will be deleted.
+
+    .PARAMETER Timeout
```
HowardWolosky, line 280: I think both `Timeout` and `RetentionPeriod` should indicate their unit of measurement in their name (`TimeoutMinutes`, `IdleRetentionPeriodMinutes`).  Would help users not to have to read the documentation to understand what the time unit is.  I also think we should keep `Idle` in the retention period name.  Users can still just use `Timeout` and `IdleRetentionPeriod` when specifying the parameters since PowerShell doesn't require full parameter names when there's no ambiguity, but I think the default for tab completion should be with `Minutes` so that users know.

HowardWolosky (changes requested):
Thanks so much for all of this work.  SUPER appreciated.
Sincere apologies as well on the delayed review.  Life happened and I couldn't give this the dedicated time it needed prior to today.

There's a lot of feedback here, but none of it is super large, so I don't think it will take too long to apply any of it (aside from some of the questions I have in some of the tests around the `super-fake-test-org`).

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,945 @@
...
+        [string] $Location,
+
+        [string] $Machine,
+
+        [switch] $NoMultipleRepoPermissions,
+
+        [ValidateRange(0, 43200)]
+        [int] $RetentionPeriod,
+
+        [ValidateRange(5, 240)]
```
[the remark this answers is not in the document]
vercellone: Yes. I had it set lower and an error is returned stating as much.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,945 @@
...
+        # Map params to hashBody properties
+        foreach ($p in $PSBoundParameters.GetEnumerator())
+        {
+            if ($propertyMap.ContainsKey($p.Key) -and -not [string]::IsNullOrWhiteSpace($p.Value))
+            {
+                $hashBody.Add($propertyMap[$p.Key], $p.Value)
+            }
+        }
+
+        if ($PSCmdlet.ParameterSetName -eq 'AuthenticatedUser')
```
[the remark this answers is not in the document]
vercellone: > This endpoint requires either a repository_id OR a pull_request but not both.

When pull_request object is included, repository_id is included as a property of that object.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,945 @@
...
+        https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#create-a-codespace-in-a-repository
+
+    .LINK
+        https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#create-a-codespace-from-a-pull-request
+#>
+    [CmdletBinding(
+        SupportsShouldProcess,
+        DefaultParameterSetName = 'AuthenticatedUser')]
+    [OutputType({ $script:GitHubCodespaceTypeName })]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSReviewUnusedParameter", "", Justification = "The Uri parameter is only referenced by Resolve-RepositoryElements which get access to it from the stack via Get-Variable -Scope 1.")]
```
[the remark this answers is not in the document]
vercellone: Not sure I follow.  In the SuppressMessageAttribute Justification?

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,945 @@
...
+            else
+            {
+                $hashBody.Add('repository_id', $RepositoryId)
+            }
+        }
+
+        $params = @{
+            UriFragment = $uriFragment
+            Body = (ConvertTo-Json -InputObject $hashBody)
+            Method = 'POST'
```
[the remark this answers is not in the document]
vercellone: https://www.rfc-editor.org/rfc/rfc7231#section-4.1 states

> By convention, standardized methods are defined in all-uppercase US-ASCII letters

So, I always make them uppercase, but will conform for the sake of consistency.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,945 @@
...
+        The Git repo for this module can be found here: http://aka.ms/PowerShellForGitHub
+
+    .PARAMETER CodespaceName
+        Name of the Codespace.
+
+    .PARAMETER Wait
+        If present will wait for the codespace to start.
+
+    .PARAMETER PassThru
+        Returns the updated GitHub Issue.  By default, this cmdlet does not generate any output.
```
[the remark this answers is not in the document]
vercellone: Returns the start/stop action result.  By default, this cmdlet does not generate any output.

I'll correct both.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,945 @@
...
+        https://docs.github.com/en/rest/codespaces/codespaces?apiVersion=2022-11-28#delete-a-codespace-for-the-authenticated-user
+
+    .LINK
+        https://docs.github.com/en/rest/codespaces/organizations?apiVersion=2022-11-28#delete-a-codespace-from-the-organization
+#>
+    [CmdletBinding(
+        DefaultParameterSetName = 'AuthenticatedUser',
+        SupportsShouldProcess,
+        ConfirmImpact = 'High')]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSReviewUnusedParameter", "", Justification = "PassThru is accessed indirectly via Resolve-ParameterWithDefaultConfigurationValue")]
```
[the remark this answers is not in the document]
vercellone: Nope

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,945 @@
...
+    $result = (Invoke-GHRestMethod @params | Add-GitHubCodespaceAdditionalProperties)
+
+    if ($Wait.IsPresent)
+    {
+        $waitParams = @{
+            CodespaceName = $CodespaceName
+            AccessToken = $AccessToken
+        }
+
+        $result = Wait-GitHubCodespaceAction @waitParams
```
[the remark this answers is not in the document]
vercellone: Noted

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,945 @@
...
+    .EXAMPLE
+        Wait-GitHubCodespace -Name vercellone-effective-goggles-qrv997q6j9929jx8
+
+     .NOTES
+        Internal-only helper method.
+#>
+    [CmdletBinding()]
+    [OutputType({ $script:GitHubCodespaceTypeName })]
+    param(
+        [Parameter(Mandatory,
```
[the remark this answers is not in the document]
vercellone: You've got an extra comma there.  But, I get it.  Fixing.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,450 @@
...
+    }
+    $repo = New-GitHubRepository @newGitHubRepositoryParms
+}
+
+Describe 'GitHubCodespaces\Delete-GitHubCodespace' {
+
+    Context 'When deleting a codespace for the authenticated user' {
+        BeforeEach {
+            # Suppress HTTP 202 warning for codespace creation
+            # TODO: Suppression is not working as intended here.
```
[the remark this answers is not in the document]
vercellone: When I run these tests I see warning messages that are not indicative of failure, so I'd rather not see them at all.

Tests/GitHubCodespaces.tests.ps1
vercellone: Valid.  The default can be changed by an admin.  Removing these 30-minute assertions.

Tests/GitHubCodespaces.tests.ps1
vercellone: It does exist as a sandbox in my GitHub Enterprise environment.  But, this should use $script:organizationName based on test config Settings.  I'll correct it.

GitHubCodespaces.ps1
vercellone: as you wish...

vercellone:
> Thanks so much for all of this work. SUPER appreciated. Sincere apologies as well on the delayed review. Life happened and I couldn't give this the dedicated time it needed prior to today.

Np.  Life is like that sometimes.

> There's a lot of feedback here, but none of it is super large, so I don't think it will take too long to apply any of it (aside from some of the questions I have in some of the tests around the `super-fake-test-org`).

Good feedback. There are 4 unresolved conversations to which I replied but committed no related changes.  Of those only the  "other parameters...dynamically referenced via the propertyMap" demands changes.  I'm hoping you can elaborate on that one.

GitHubCodespaces.ps1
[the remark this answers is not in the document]
vercellone: I think I have an idea of what you are getting at.  This SuppressMessageAttribute is effective for all of those parameters mapped to properties also.  Would you like one long justification?  An attribute per parameter?  1 attribute for Uri and 1 for the mapped params?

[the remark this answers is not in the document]
vercellone: I committed a PSReviewUnusedParameter w/CheckId for each individual parameter.  This works, but let me know if you had something else in mind.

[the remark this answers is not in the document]
HowardWolosky: Sorry for not being more clear here.  I just meant something like:

```powershell
    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSReviewUnusedParameter", "", Justification = "The Uri parameter is only referenced by Resolve-RepositoryElements which get access to it from the stack via Get-Variable -Scope 1, and most of the others get dynamically accessed via `$propertyMap.")]
```

GitHubCodespaces.ps1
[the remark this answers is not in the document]
HowardWolosky: Ah --- you're totally right.  I was looking at the schema itself which called out `required` on both, but failed to read the comments just above the schema.  My bad.

GitHubCodespaces.ps1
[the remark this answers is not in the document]
vercellone: Done.

vercellone:
@HowardWolosky Are you waiting on me for anything?  What is the next step?

vercellone:
@HowardWolosky?

HowardWolosky:
Ack -- sorry about that.  Other things came up and I missed your previous ping on this.  I have this queued to look at this weekend.

HowardWolosky:
> Ack -- sorry about that. Other things came up and I missed your previous ping on this. I have this queued to look at this weekend.

Sincerest apologies here.  Clearly last weekend didn't happen.  I don't want to make another promise that can't be kept, so I'll simply promise that I'll complete my review of the last iteration by the end of _next_ weekend.  If any additional updates are required coming out of that, I'll promise a 48 hour review turnaround on those.  I greatly appreciate your contribution and want you to feel like it is valued (because it is)...I'm just trying to balance a number of different obligations here.

GitHubCodespaces.ps1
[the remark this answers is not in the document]
vercellone: Thought I had changed these already.  Done now.  Anything else?

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,443 @@
...
+                if ($codespace)
+                {
+                    Remove-GitHubCodespace -CodespaceName $codespace.name -Confirm:$false
+                }
+            }
+        }
+
+        Context -Name 'When creating a codespace with default settings with Ref' {
+            BeforeAll {
+                $repoWithPR = Get-GitHubRepository -OrganizationName $script:organizationName |
```
HowardWolosky, line 232: This one is going to fail when run against the default test accounts, as they have no repos in them when they run (unless there were failures in previous runs resulting in an incomplete cleanup, and then even if they did, no guarantee that they have any PR's).  You can make it a bit more robust by ensuring that there is a specific repo in there with a known PR:

1. Creating a new repo (`New-GitHubRepository`)
2. Creating a new branch in that repo (`New-GitHubRepositoryBranch`)
3. Making a change to the test branch (`Set-GitHubContent`)
4. Creating a new PR for that branch against the main branch (`New-GitHubPullRequest`)

```powershell
$mainBranchName = 'master'
$prBranchName = 'test'
$repoWithPR = New-GitHubRepository -RepositoryName ([guid]::NewGuid()) -AutoInit
$prBranch = $repoWithPR | New-GitHubRepositoryBranch -BranchName $mainBranchName -TargetBranchName $prBranchName
$repoWithPR | Set-GitHubContent -Path 'README.md' -CommitMessage 'test' -Content 'New content' -BranchName $prBranchName
$pullRequest = $branch | New-GitHubPullRequest -Title 'testPR' -Head $prBranchName -Base $mainBranchName -Verbose
```

> Just be sure to clean it up (`Remove-GitHubRepository`) when you're all done with it.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,443 @@
...
+                if ($codespace)
+                {
+                    Remove-GitHubCodespace -CodespaceName $codespace.name -Confirm:$false
+                }
+            }
+        }
+
+        Context -Name 'When creating a codespace with default settings from a PullRequest' {
+            BeforeAll {
+                  $repoWithPR = Get-GitHubRepository -OrganizationName $script:organizationName |
```
HowardWolosky, line 266: nit: indentation is off. 
Additionally, same feedback as above about creating the repo that has a PR in it, as opposed to expecting there to already be one.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,443 @@
...
+                {
+                    Remove-GitHubCodespace -CodespaceName $codespace.name -Confirm:$false
+                }
+            }
+        }
+
+        Context -Name 'When creating a codespace with all possible settings' {
+            BeforeAll {
+                $newGitHubCodespaceParms = @{
+                    # ClientIp = 'TODO ???? - should be instead of rather than in addition to Geo, perhaps add some param validation to the function'
```
HowardWolosky, line 301: Yeah -- this one is just too challenging to test.  I wouldn't worry about it.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,443 @@
...
+                    # DevContainerPath = 'Will add to test in the future when Get-GitHubDevContainer is implemented and the test repo includes one'
+                    DisplayName = 'PowerShellForGitHub pester test'
+                    Geo = 'UsWest'
+                    Machine = 'basicLinux32gb'
+                    NoMultipleRepoPermissions = $true # Not sure how to assert this, but this proves it accepts the switch without error
+                    IdleRetentionPeriodMinutes = 10
+                    TimeoutMinutes = 5
+                    # WorkingDirectory = 'TODO ???? - not sure how to handle this'
```
HowardWolosky, line 309: The real way to do it would be to mock out `Invoke-GHRestMethod` and just validate that the JSON body has the expected value, but all of the other tests that we do here are full end-to-end tests against the real API.  So, I wouldn't worry about this one either.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        Context -Name 'When creating a codespace with all possible settings' {
+            BeforeAll {
+                $newGitHubCodespaceParms = @{
+                    # ClientIp = 'TODO ???? - should be instead of rather than in addition to Geo, perhaps add some param validation to the function'
+                    # DevContainerPath = 'Will add to test in the future when Get-GitHubDevContainer is implemented and the test repo includes one'
+                    DisplayName = 'PowerShellForGitHub pester test'
+                    Geo = 'UsWest'
+                    Machine = 'basicLinux32gb'
```
HowardWolosky, line 305: I was looking for where these names came from.  Eventually found [this](https://docs.github.com/en/rest/codespaces/machines?apiVersion=2022-11-28):

```powershell
(Invoke-GHRestMethod -UriFragment '/repos/PowerShellForGitHubTeam/6cd3a09f-9275-45a6-b908-f0e6b294afcf/codespaces/machines' -Method Get -Description 'Get available Codespace machine types').machines
```

We should definitely get your current change checked-in (once these remaining review comments are resolved), but a TODO in my head is to get those two API's added as well so that users can find out what possible machines can be used when using this module.

HowardWolosky (changes requested):
Thanks for your patience on this updated review.
I've added some new feedback for you to consider, and re-activated some previous comments that had been resolved without any corresponding change (or alternatively a corresponding comment explaining why there was no change in response).

I'll run the tests locally after the next update to make sure that nothing was missed.

Thanks again for this contribution (and again, for your patience)!

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,950 @@
...
+    .PARAMETER DevContainerPath
+        Path to devcontainer.json config to use for this codespace.
+
+    .PARAMETER DisplayName
+        Display name for this codespace
+
+    .PARAMETER Geo
+        The geographic area for this codespace.
+        Assigned by IP if not provided.
+        Can be one of: EuropeWest, SoutheastAsia, UsEast, UsWest
```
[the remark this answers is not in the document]
vercellone: This seems risky.  I deliberately avoided it because geo is relatively new.  I'm worried the ValidateSet might be debilitatingly restrictive should they change or add options.  But, I suppose it isn't terribly difficult to change if needed.

GitHubCodespaces.ps1
```diff
@@ -0,0 +1,950 @@
...
+    .PARAMETER IdleRetentionPeriodMinutes
+        Duration in minutes (up to 30 days) after codespace has gone idle in which it will be deleted.
+
+    .PARAMETER TimeoutMinutes
+        Time in minutes before codespace stops from inactivity.
+
+    .PARAMETER WorkingDirectory
+        Working directory for this codespace.
+
+    .PARAMETER AccessToken
```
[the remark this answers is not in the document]
vercellone: sure.

Tests/GitHubCodespaces.tests.ps1
vercellone: Reasonable. Will try to remember to add both [machines](https://docs.github.com/en/rest/codespaces/machines?apiVersion=2022-11-28) endpoints in the next PR.  These would be good candidates for ArgumentCompleters, but I see you don't have any - probably due to compatibility concerns.

Tests/GitHubCodespaces.tests.ps1
vercellone: Common Before/After All creates and destroys a $repo, so I've reused that but did create a new branch and PR as suggested.

My default test org's default branch is main, so I opted for dynamicly retrieving with `$mainBranchName = $repo | Get-GitHubRepositoryBranch | Select-Object -ExpandProperty name -First 1`.  Let me know if there's a better way.

vercellone:
> Thanks for your patience on this updated review. I've added some new feedback for you to consider, and re-activated some previous comments that had been resolved without any corresponding change (or alternatively a corresponding comment explaining why there was no change in response).
> 
> I'll run the tests locally after the next update to make sure that nothing was missed.
> 
> Thanks again for this contribution (and again, for your patience)!

Back to you @HowardWolosky.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,459 @@
...
+            $codespaces = Get-GitHubCodespace |
+                Where-Object { $_.repository.name -eq $defaultRepositoryName }
+        }
+
+        It 'Should return objects of the correct type' {
+            $codespaces[0].PSObject.TypeNames[0] | Should -Be 'GitHub.Codespace'
+        }
+
+        It 'Should return one or more results' {
+            $codespaces.Count | Should -BeGreaterOrEqual 1
```
HowardWolosky, line 90: This is failing.  `$codespaces` is coming back as a `PSCustomObject`, not a `PSCustomObject[]`, so there is no `Count` on it.

``
 Expected the actual value to be greater than or equal to 1, but got $null.
``

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,459 @@
...
+            {
+                $codespace.owner.login | Should -Be $script:OwnerName
+                $codespace.repository.name | Should -Be $newGitHubCodespaceParms.RepositoryName
+            }
+        }
+    }
+
+    Context 'When getting a codespace for a specified owner and repository' {
+        BeforeAll {
+            $codespaces = Get-GitHubCodespace @newGitHubCodespaceParms
```
HowardWolosky, line 104: `Get-GitHubCodespace` doesn't have `Wait`, but `$newGitHubCodespaceParms` does.  That results in this failure:

``
[-] Context GitHubCodespaces\Get-GitHubCodespace.When getting a codespace for a specified owner and repository failed
 ParameterBindingException: A parameter cannot be found that matches parameter name 'Wait'. This failure might be caused by applying the default parameter binding. You can disable the default parameter binding in $PSDefaultParameterValues by setting $PSDefaultParameterValues["Disabled"] to be $true, and retry. The following default parameter was successfully bound for this cmdlet when the error occurred: -OutVariable
 at <ScriptBlock>, D:\Source\PowerShellForGitHub\tests\GitHubCodespaces.tests.ps1:104
``

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,459 @@
...
+            {
+                $codespace.owner.login | Should -Be $script:OwnerName
+                $codespace.repository.name | Should -Be $newGitHubCodespaceParms.RepositoryName
+            }
+        }
+    }
+
+    Context 'When getting all codespaces for a specified organization' {
+        BeforeAll {
+            $codespaces = Get-GitHubCodespace -OrganizationName $script:organizationName
```
HowardWolosky, line 126: This is `$null` because within the context of this test, no codespaces have been created yet.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,459 @@
...
+        }
+
+        It 'Should return objects of the correct type' {
+            $codespaces[0].PSObject.TypeNames[0] | Should -Be 'GitHub.Codespace'
+        }
+
+        It 'Should return one or more results' {
+            $codespaces.Count | Should -BeGreaterOrEqual 1
```
HowardWolosky, line 134: Same issue as at 126.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,459 @@
...
+            OwnerName = $script:organizationName
+            RepositoryName = $defaultRepositoryName
+            Wait = $true
+        }
+        $null = New-GitHubCodespace @newGitHubCodespaceParms
+    }
+
+    Context 'When starting a codespace for the authenticated user' {
+        BeforeAll {
+            $codespace = Get-GitHubCodespace @newGitHubCodespaceParms
```
HowardWolosky, line 399: Same issue as at 104.

HowardWolosky (changes requested):
This is looking really close.  Thanks for the quick updates.

Found just a couple lingering things in the implementation.  Most of the feedback is in the tests, as I finally ran them locally and hit a bunch of problems. I've tried noting the exact issues directly as comments.  Please go through another local run of the tests to ensure they're clean before your next update.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,459 @@
...
+
+            It 'Should return an object of the correct type' {
+                $codespace | Should -BeOfType PSCustomObject
+            }
+
+            It 'Should return the correct properties' {
+                # $codespace.devcontainer_path | Should -Be
+                $codespace.display_name | Should -Be $newGitHubCodespaceParms.DisplayName
+                $codespace.idle_timeout_minutes | Should -Be $newGitHubCodespaceParms.TimeoutMinutes
+                $codespace.location | Should -Match 'WestUs'
```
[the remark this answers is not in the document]
vercellone: I disagree.  We request Geo=UsWest, and the returned location should reflect that.  It should not return a EastUs location.  And, there is no Geo property in the output.

Tests/GitHubCodespaces.tests.ps1
vercellone: I'm referring to https://learn.microsoft.com/en-us/previous-versions/powershell/module/microsoft.powershell.core/about/about_functions_argument_completion?view=powershell-7.1.  I think it was officially introduced in PowerShell 5, but I also think there is a module which provides Register-ArgumentCompleter support for older versions.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,459 @@
...
+    BeforeAll {
+        # Suppress HTTP 202 warning for codespace creation
+        $WarningPreference = 'SilentlyContinue'
+
+        $newGitHubCodespaceParms = @{
+            OwnerName = $script:organizationName
+            RepositoryName = $defaultRepositoryName
+            Wait = $true
+        }
+        $null = New-GitHubCodespace @newGitHubCodespaceParms
```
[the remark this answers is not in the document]
vercellone: Fixed with More aggressive AfterAll Remove-GitHubCodespace commit e5ae3e9

Tests/GitHubCodespaces.tests.ps1
vercellone: This test passes for me because the repository of the codespace created in the BeforeAll is in the same Organization (Settings $script:organizationName).  I am not sure how to guarantee that in both mine and your environment.  If you have an idea please share.  If not, perhaps it is best to drop this test altogether.

vercellone:
**2 left**
1. [Should we test _When getting all codespaces for a specified organization_?  If so, how?](https://github.com/microsoft/PowerShellForGitHub/pull/407/files#r1301955717)
1. [Should we drop the location assertion? I vote no.](https://github.com/microsoft/PowerShellForGitHub/pull/407/files#r1301942956)

Let me know.

Tests/GitHubCodespaces.tests.ps1
vercellone: I'm going to remove Context 'When getting all codespaces for a specified organization', for now.  If you insist on maintaining it, then I'll need help implementing it in a way that works in your environment.

HowardWolosky: > This test passes for me because the repository of the codespace created in the BeforeAll is in the same Organization (Settings $script:organizationName). I am not sure how to guarantee that in both mine and your environment. If you have an idea please share. If not, perhaps it is best to drop this test altogether.

The issue is that you haven't in the context of this test, you haven't created a codespace yet.  In the root `BeforeAll`  (lines 15-35) you only created a repo.  That's the _only_ guaranteed thing that you have at the moment.  In 125-127, you're querying for codespaces within that org.  If you are expecting that to return back a non-zero number of codespaces, then it's up to you to ensure that you create a codespace within the context of that org (in that repo at 32?) before querying for the org's codespaces.

Tests/GitHubCodespaces.tests.ps1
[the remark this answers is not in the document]
HowardWolosky: I missed both that you were setting the `Geo` in this test, as well as the fact that `Geo` isn't in returned response.  I'm ok with you keeping this then, but it may be worth just adding a comment to 323 that references the fact that this is the translated location based on the specified `USWest` Geo.

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,450 @@
...
+            $codespaces = Get-GitHubCodespace |
+                Where-Object { $_.repository.name -eq $defaultRepositoryName }
+        }
+
+        It 'Should return objects of the correct type' {
+            $codespaces[0].PSObject.TypeNames[0] | Should -Be 'GitHub.Codespace'
+        }
+
+        It 'Should return one or more results' {
+            ($codespaces | Where-Object { $_ }).Count | Should -BeGreaterOrEqual 1
```
HowardWolosky, line 88: ``
[-] GitHubCodespaces\Get-GitHubCodespace.When getting codespaces for the authenticated user.Should return one or more results 55ms (52ms|3ms)
 Expected the actual value to be greater than or equal to 1, but got $null.
 at ($codespaces | Where-Object { $_ }).Count | Should -BeGreaterOrEqual 1, C:\Source\PowerShellForGitHub\tests\GitHubCodespaces.tests.ps1:88
 at <ScriptBlock>, C:\Source\PowerShellForGitHub\tests\GitHubCodespaces.tests.ps1:88
``

Tests/GitHubCodespaces.tests.ps1
```diff
@@ -0,0 +1,450 @@
...
+    Context 'When getting a codespace for a specified owner and repository' {
+        BeforeAll {
+            $codespaces = Get-GitHubCodespace @newGitHubCodespaceParms
+        }
+
+        It 'Should return objects of the correct type' {
+            $codespaces[0].PSObject.TypeNames[0] | Should -Be 'GitHub.Codespace'
+        }
+
+        It 'Should return one or more results' {
```
HowardWolosky, line 109: ``
[-] GitHubCodespaces\Get-GitHubCodespace.When getting a codespace for a specified owner and repository.Should return one or more results 13ms (11ms|2ms)
 Expected the actual value to be greater than or equal to 1, but got $null.
 at ($codespaces | Where-Object { $_ }).Count | Should -BeGreaterOrEqual 1, C:\Source\PowerShellForGitHub\tests\GitHubCodespaces.tests.ps1:110
 at <ScriptBlock>, C:\Source\PowerShellForGitHub\tests\GitHubCodespaces.tests.ps1:110
``

Tests/GitHubCodespaces.tests.ps1
HowardWolosky: Ah, thanks.  We were looking at the same docs, just different ways of accomplishing it.  The `IValidateSetValuesGenerator` approach requires `Classes` which weren't introduced until PS6.  It looks like the `ArgumentCompletions` attribute was made available in PS5.  I'm not quite sure how I feel about having to do a network call for tab completion to work.  While nifty, I think that approach is likely overkill for our immediate needs, since an invalid value will likely get a reasonable error message from the API.  I think it should be enough for there to be an available API for uses to call to know what possible values could be used for that.  But again, this is something that can be hashed out further in a future PR.

HowardWolosky (changes requested):
Looks like you still have a couple failing tests.

Tests/GitHubCodespaces.tests.ps1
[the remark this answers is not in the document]
vercellone: Comments added.

Tests/GitHubCodespaces.tests.ps1
vercellone: Agreed.

Tests/GitHubCodespaces.tests.ps1
vercellone: Should be resolved now with `OwnerName = $repo.owner.login`.

Tests/GitHubCodespaces.tests.ps1
vercellone: Should be resolved now with `OwnerName = $repo.owner.login`.

vercellone:
```PowerShell
Invoke-Pester -Configuration (
    [PesterConfiguration]@{
        Filter = @{ FullName = 'GitHubCodespaces\*-GitHubCodespace' }
        Output = @{ Verbosity = 'Detailed' }
        Path   = 'C:\Users\Jason\source\GitHub\PowerShellForGitHub\Tests\GitHubCodespaces.tests.ps1'
    }
)

Pester v5.4.1

Starting discovery in 28 files.
Discovery found 1128 tests in 616ms.
Filter 'FullName' set to ('GitHubCodespaces\*-GitHubCodespace').
Filters selected 30 tests to run.
Running tests.

Running tests from 'C:\Users\Jason\source\GitHub\PowerShellForGitHub\Tests\GitHubCodespaces.tests.ps1'
Describing GitHubCodespaces\Delete-GitHubCodespace
 Context When deleting a codespace for the authenticated user
   [+] Should get no content using -Confirm:$false 15.06s (15.06s|2ms)
   [+] Should get no content using -Force 15.02s (15.02s|1ms)

Describing GitHubCodespaces\Get-GitHubCodespace
 Context When getting codespaces for the authenticated user
   [+] Should return objects of the correct type 3ms (2ms|2ms)
   [+] Should return one or more results 3ms (2ms|1ms)
   [+] Should return the correct properties 2ms (2ms|1ms)
 Context When getting a codespace for a specified owner and repository
   [+] Should return objects of the correct type 3ms (2ms|1ms)
   [+] Should return one or more results 3ms (2ms|1ms)
   [+] Should return the correct properties 3ms (2ms|1ms)
 Context When getting a codespace for a specified organization user
   [+] Should have results for the organization user 4ms (3ms|1ms)
   [+] Should return the correct properties 3ms (2ms|1ms)
 Context When getting a codespace for a specified codespace name
   [+] Should return objects of the correct type 3ms (2ms|1ms)
   [+] Should return the correct properties 2ms (2ms|1ms)
 Context When specifiying the Uri parameter
   [+] Should return objects of the correct type 3ms (2ms|1ms)
   [+] Should return the correct properties 3ms (2ms|1ms)
 Context When specifiying the Uri parameter from the pipeline
   [+] Should return objects of the correct type 4ms (2ms|1ms)
   [+] Should return the correct properties 2ms (2ms|1ms)

Describing GitHubCodespaces\New-GitHubCodespace
 Context When creating a repository for the authenticated user
  Context When creating a codespace with default settings with RepositoryId
    [+] Should return an object of the correct type 10ms (6ms|3ms)
    [+] Should return the correct properties 4ms (3ms|1ms)
  Context When creating a codespace with default settings with Ref
    [+] Should return an object of the correct type 4ms (2ms|2ms)
    [+] Should return the correct properties 4ms (3ms|1ms)
  Context When creating a codespace with default settings from a PullRequest
    [+] Should return an object of the correct type 3ms (2ms|1ms)
    [+] Should return the correct properties 4ms (3ms|1ms)
  Context When creating a codespace with all possible settings
    [+] Should return an object of the correct type 3ms (2ms|1ms)
    [+] Should return the correct properties 7ms (7ms|1ms)
  Context When creating a codespace with default settings with Repository Elements
    [+] Should return an object of the correct type 3ms (2ms|1ms)
    [+] Should return the correct properties 4ms (3ms|1ms)

Describing GitHubCodespaces\Start-GitHubCodespace
 Context When starting a codespace for the authenticated user
   [+] Should not throw 4.26s (4.25s|2ms)
   [+] Should become Available 7.48s (7.48s|1ms)

Describing GitHubCodespaces\Stop-GitHubCodespace
 Context When stopping a codespace for the authenticated user
   [+] Should not throw 3.9s (3.9s|3ms)
   [+] Should become Shutdown 6.78s (6.78s|1ms)
Tests completed in 206.51s
Tests Passed: 30, Failed: 0, Skipped: 0 NotRun: 1098
```

HowardWolosky (changes requested):
Still seeing two test failures:

``
[-] GitHubCodespaces\Get-GitHubCodespace.When getting codespaces for the authenticated user.Should return one or more results 44ms (42ms|2ms)
 Expected the actual value to be greater than or equal to 1, but got $null.
 at ($codespaces | Where-Object { $_ }).Count | Should -BeGreaterOrEqual 1, C:\Source\PowerShellForGitHub\tests\GitHubCodespaces.tests.ps1:88
 at <ScriptBlock>, C:\Source\PowerShellForGitHub\tests\GitHubCodespaces.tests.ps1:88
[-] GitHubCodespaces\Get-GitHubCodespace.When getting a codespace for a specified owner and repository.Should return one or more results 12ms (11ms|1ms)
 Expected the actual value to be greater than or equal to 1, but got $null.
 at ($codespaces | Where-Object { $_ }).Count | Should -BeGreaterOrEqual 1, C:\Source\PowerShellForGitHub\tests\GitHubCodespaces.tests.ps1:110
 at <ScriptBlock>, C:\Source\PowerShellForGitHub\tests\GitHubCodespaces.tests.ps1:110
``

vercellone:
> Still seeing two test failures:

Thanks for the quick responses this week.

I could not reproduce these last 2 test failures in pwsh 7.3.6, but I was able to reproduce them in Windows PowerShell 5.1.  They require an @ symbol before the ($codespaces...).Count to force the result to an Object[] array and avoid the InvokeMethodOnNull exception.

I think we should be good now.

Formatters/GitHubCodespaces.Format.ps1xml
```diff
@@ -0,0 +1,41 @@
...
+                <PropertyName>name</PropertyName>
+              </TableColumnItem>
+              <TableColumnItem>
+                <PropertyName>display_name</PropertyName>
+              </TableColumnItem>
+              <TableColumnItem>
+                <PropertyName>state</PropertyName>
+              </TableColumnItem>
+              <TableColumnItem>
+                <PropertyName>geo</PropertyName>
```
HowardWolosky, line 27: If `geo` isn't in the response data, we probably don't want to have this in our formatter as this column then will always be blank.

HowardWolosky:
Can you also add `last_used_at` to this (will ensure that it gets converted to a real date/time object in the return object by default)?

https://github.com/microsoft/PowerShellForGitHub/blob/482fe2331732a24ed779b140c3f38c8a156271e1/GitHubCore.ps1#L990-L1008

vercellone:
last_used_at added to datePropertyNames

HowardWolosky (approved):
Thanks so much for all your hard work in getting this support added.  Very much appreciated.  Looking forward to hitting the button to merge this.

PR #444, opened 2025-01-08
Added TLS 1.3 support and fixed strict-mode variable errors in tests

### LINKED ISSUE #443: Only the TLS 1.2 protocol is supported
#### Feature Idea Summary
PowerShell Core supports the faster TLS 1.3 protocol, but the module forces it to use 1.2.

#### Feature Idea Additional Details
In the GitHubCore.ps1 file:
```powershell
[Net.ServicePointManager]::SecurityProtocol=[Net.SecurityProtocolType]::Tls12
```

#### Requested Assignment
Don't care. I can do it if the PR doesn't take too long. Just change the above code to:
```powershell
if ($PSVersionTable.PSVersion -lt 7.0.0)
{
    [Net.ServicePointManager]::SecurityProtocol=[Net.SecurityProtocolType]::Tls12
}
```
and in the finally block:
```powershell
if ($PSVersionTable.PSVersion -lt 7.0.0)
{
    [Net.ServicePointManager]::SecurityProtocol = $originalSecurityProtocol
}
```

PowerShell 6 isn't supported anymore, so anything less than 7 should use TLS 1.2.

#### Operating System
All

#### PowerShell Version
```
PSVersion                      7.4.6
GitCommitId                    7.4.6
OS                             Microsoft Windows 10.0.19045
```

#### Module Version
Running: 0.17.0
Installed: 0.17.0

variableresistor:
#### Description
Added support for TLS 1.3

#### Issues Fixed
Fixes #443 

#### Checklist

- [x] [Formatters were created](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#formatters) for any new types being added.
- [x] Unit tests were added/updated and are all passing. See [testing guidelines](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#testing).  This includes making sure that all pipeline input variations have been covered.
- [x] If desired, ensure your name is added to our [Contributors list](https://github.com/microsoft/PowerShellForGitHub/blob/master/CONTRIBUTING.md#contributors)

### DIFF
--- a/GitHubCore.ps1
+++ b/GitHubCore.ps1
@@ -317,7 +317,7 @@ function Invoke-GHRestMethod
 
             if ($PSVersionTable.PSVersion -lt [version]"7.0.0")
             {
-                [Net.ServicePointManager]::SecurityProtocol=[Net.SecurityProtocolType]::Tls12
+                [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
             }
 
             $result = Invoke-WebRequest @params
--- a/GitHubCore.ps1
+++ b/GitHubCore.ps1
@@ -26,6 +26,7 @@
  }
 
 Set-Variable -Scope Script -Option ReadOnly -Name ValidBodyContainingRequestMethods -Value ('Post', 'Patch', 'Put', 'Delete')
+Set-Variable -Scope Script -Option ReadOnly -Name PreferredSecurityProtocolType -Value (Get-PreferredSecurityProtocolType)
 
 function Invoke-GHRestMethod
 {
@@ -275,6 +276,7 @@ function Invoke-GHRestMethod
     }
 
     $originalSecurityProtocol = [Net.ServicePointManager]::SecurityProtocol
+    [Net.ServicePointManager]::SecurityProtocol = $PreferredSecurityProtocolType
 
     # When $Save is in use, we need to remember what file we're saving the result to.
     $outFile = [String]::Empty
@@ -314,11 +316,6 @@ function Invoke-GHRestMethod
 
             # Disable Progress Bar in function scope during Invoke-WebRequest
             $ProgressPreference = 'SilentlyContinue'
-
-            if ($PSVersionTable.PSVersion -lt [version]"7.0.0")
-            {
-                [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
-            }
 
             $result = Invoke-WebRequest @params
 
@@ -585,10 +582,7 @@ function Invoke-GHRestMethod
     }
     finally
     {
-        if ($PSVersionTable.PSVersion -lt [Version]'7.0.0')
-        {
-            [Net.ServicePointManager]::SecurityProtocol = $originalSecurityProtocol
-        }
+        [Net.ServicePointManager]::SecurityProtocol = $originalSecurityProtocol
     }
 }
 
--- a/GitHubCore.ps1
+++ b/GitHubCore.ps1
@@ -315,7 +315,7 @@ function Invoke-GHRestMethod
             # Disable Progress Bar in function scope during Invoke-WebRequest
             $ProgressPreference = 'SilentlyContinue'
 
-            if ($PSVersionTable.PSVersion -lt 7.0.0)
+            if ($PSVersionTable.PSVersion -lt [version]"7.0.0")
             {
                 [Net.ServicePointManager]::SecurityProtocol=[Net.SecurityProtocolType]::Tls12
             }
@@ -585,7 +585,7 @@ function Invoke-GHRestMethod
     }
     finally
     {
-        if ($PSVersionTable.PSVersion -lt 7.0.0)
+        if ($PSVersionTable.PSVersion -lt [version]"7.0.0")
         {
             [Net.ServicePointManager]::SecurityProtocol = $originalSecurityProtocol
         }
--- a/GitHubCore.ps1
+++ b/GitHubCore.ps1
@@ -315,7 +315,10 @@ function Invoke-GHRestMethod
             # Disable Progress Bar in function scope during Invoke-WebRequest
             $ProgressPreference = 'SilentlyContinue'
 
-            [Net.ServicePointManager]::SecurityProtocol=[Net.SecurityProtocolType]::Tls12
+            if ($PSVersionTable.PSVersion -lt 7.0.0)
+            {
+                [Net.ServicePointManager]::SecurityProtocol=[Net.SecurityProtocolType]::Tls12
+            }
 
             $result = Invoke-WebRequest @params
 
@@ -582,7 +585,10 @@ function Invoke-GHRestMethod
     }
     finally
     {
-        [Net.ServicePointManager]::SecurityProtocol = $originalSecurityProtocol
+        if ($PSVersionTable.PSVersion -lt 7.0.0)
+        {
+            [Net.ServicePointManager]::SecurityProtocol = $originalSecurityProtocol
+        }
     }
 }
 
--- a/GitHubCore.ps1
+++ b/GitHubCore.ps1
@@ -585,7 +585,7 @@ function Invoke-GHRestMethod
     }
     finally
     {
-        if ($PSVersionTable.PSVersion -lt [version]"7.0.0")
+        if ($PSVersionTable.PSVersion -lt [Version]'7.0.0')
         {
             [Net.ServicePointManager]::SecurityProtocol = $originalSecurityProtocol
         }
--- a/Helpers.ps1
+++ b/Helpers.ps1
@@ -709,3 +709,35 @@ function New-ErrorRecord
 
     return $errorRecord
 }
+
+filter Get-PreferredSecurityProtocolType
+{
+    <#
+    .SYNOPSIS
+        Decides the security protocol to use when making a REST API call
+
+    .DESCRIPTION
+        Decides to use TLS 1.3 if the user is running on PowerShell version 7.0 or higher, otherwise
+        uses TLS 1.2 if the user is running on Windows PowerShell.
+
+    .NOTES
+        See https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#minimum-version
+        TLS Best Practices https://learn.microsoft.com/en-us/dotnet/framework/network-programming/tls
+    #>
+    [CmdletBinding()]
+    [OutputType([Net.SecurityProtocolType])]
+    param()
+    $dotNet47 = 460798
+    if (($PSVersionTable.PSVersion -ge [version]'7.0.0') -or (
+        (Get-ItemPropertyValue -LiteralPath 'HKLM:SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full' -Name Release) -ge $dotNet47)
+    )
+    {
+        # .NET 4.7 (460798) is the first .NET version to respect the system default and supports TLS 1.3 out-of-box
+        return [Net.SecurityProtocolType]::SystemDefault
+    }
+    elseif (([Net.SecurityProtocolType] | Get-Member -MemberType Property -Static).Name -contains 'Tls13')
+    {
+        return [Net.SecurityProtocolType]::Tls13
+    }
+    else { return [Net.SecurityProtocolType]::Tls12 }
+}
--- a/Tests/GitHubBranches.tests.ps1
+++ b/Tests/GitHubBranches.tests.ps1
@@ -322,7 +322,7 @@ Describe 'GitHubBranches\New-GitHubRepositoryBranch' {
     }
 
     AfterAll -ScriptBlock {
-        if (Get-Variable -Name repo -ErrorAction Ignore)
+        if (Get-Variable -Name repo -ErrorAction SilentlyContinue)
         {
             Remove-GitHubRepository -Uri $repo.svn_url -Confirm:$false
         }
@@ -408,7 +408,7 @@ Describe 'GitHubBranches\Remove-GitHubRepositoryBranch' {
     }
 
     AfterAll -ScriptBlock {
-        if (Get-Variable -Name repo -ErrorAction Ignore)
+        if (Get-Variable -Name repo -ErrorAction SilentlyContinue)
         {
             Remove-GitHubRepository -Uri $repo.svn_url -Confirm:$false
         }
@@ -470,7 +470,7 @@ Describe 'GitHubBranches\Get-GitHubRepositoryBranchProtectionRule' {
         }
 
         AfterAll -ScriptBlock {
-            if (Get-Variable -Name repo -ErrorAction Ignore)
+            if (Get-Variable -Name repo -ErrorAction SilentlyContinue)
             {
                 Remove-GitHubRepository -Uri $repo.svn_url -Confirm:$false
             }
@@ -723,7 +723,7 @@ Describe 'GitHubBranches\New-GitHubRepositoryBranchProtectionRule' {
     }
 
     AfterAll -ScriptBlock {
-        if (Get-Variable -Name repo -ErrorAction Ignore)
+        if (Get-Variable -Name repo -ErrorAction SilentlyContinue)
         {
             Remove-GitHubRepository -Uri $repo.svn_url -Confirm:$false
         }
@@ -785,7 +785,7 @@ Describe 'GitHubBranches\Remove-GitHubRepositoryBranchProtectionRule' {
 
     }
     AfterAll -ScriptBlock {
-        if (Get-Variable -Name repo -ErrorAction Ignore)
+        if (Get-Variable -Name repo -ErrorAction SilentlyContinue)
         {
             Remove-GitHubRepository -Uri $repo.svn_url -Confirm:$false
         }
@@ -975,7 +975,7 @@ Describe 'GitHubBranches\Get-GitHubRepositoryBranchPatternProtectionRule' {
     }
 
     AfterAll -ScriptBlock {
-        if (Get-Variable -Name repo -ErrorAction Ignore)
+        if (Get-Variable -Name repo -ErrorAction SilentlyContinue)
         {
             $repo | Remove-GitHubRepository -Force
         }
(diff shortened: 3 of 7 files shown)


HowardWolosky (changes requested):
@variableresistor - This is a great idea.  I didn't realize that PS Core 7+ supported TLS 1.3.

That being said, I'm not confident with the change.

I opened up a Windows PowerShell console:

```powershell
C:\Users\howard> $PSVersionTable.PSVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      26100  2161

C:\Users\howard> $PSVersionTable.PSVersion -lt 7.0.0
False


C:\Users\howard> $PSVersionTable.PSVersion -lt 5.01.26100
False
C:\Users\howard> $PSVersionTable.PSVersion -lt 5.01.26101
False
C:\Users\howard> $PSVersionTable.PSVersion -lt 5.1.26101
False
C:\Users\howard> $PSVersionTable.PSVersion -lt 5.1.26100
False
```

[Looking online](https://stackoverflow.com/questions/48424152/compare-system-version-in-powershell), it looks like there's more that needs to be done than what you currently have in this PR.

variableresistor:
> @variableresistor - This is a great idea. I didn't realize that PS Core 7+ supported TLS 1.3.
> 
> That being said, I'm not confident with the change.
> 
> I opened up a Windows PowerShell console:
> 
> ```powershell
> C:\Users\howard> $PSVersionTable.PSVersion
> 
> Major  Minor  Build  Revision
> -----  -----  -----  --------
> 5      1      26100  2161
> 
> C:\Users\howard> $PSVersionTable.PSVersion -lt 7.0.0
> False
> 
> 
> C:\Users\howard> $PSVersionTable.PSVersion -lt 5.01.26100
> False
> C:\Users\howard> $PSVersionTable.PSVersion -lt 5.01.26101
> False
> C:\Users\howard> $PSVersionTable.PSVersion -lt 5.1.26101
> False
> C:\Users\howard> $PSVersionTable.PSVersion -lt 5.1.26100
> False
> ```
> 
> [Looking online](https://stackoverflow.com/questions/48424152/compare-system-version-in-powershell), it looks like there's more that needs to be done than what you currently have in this PR.

Thanks @HowardWolosky for the quick reply. I'll check that link and get back with you shortly.

variableresistor:
Fixed it. To test:
```powershell
powershell -Command '$PSVersionTable.PSVersion -lt [version]"7.0.0"'
True
pwsh -Command '$PSVersionTable.PSVersion -lt [version]"7.0.0"'
False
```

GitHubCore.ps1
```diff
@@ -315,7 +315,10 @@ function Invoke-GHRestMethod
             # Disable Progress Bar in function scope during Invoke-WebRequest
             $ProgressPreference = 'SilentlyContinue'
 
-            [Net.ServicePointManager]::SecurityProtocol=[Net.SecurityProtocolType]::Tls12
+            if ($PSVersionTable.PSVersion -lt [version]"7.0.0")
```
```suggestion
            if ($PSVersionTable.PSVersion -lt [Version]'7.0.0')
```

Prefer single-quotes for non-interpreted strings, and appropriately capitalizing type names

GitHubCore.ps1
```diff
@@ -582,7 +585,10 @@ function Invoke-GHRestMethod
     }
     finally
     {
-        [Net.ServicePointManager]::SecurityProtocol = $originalSecurityProtocol
+        if ($PSVersionTable.PSVersion -lt [version]"7.0.0")
```
```suggestion
        if ($PSVersionTable.PSVersion -lt [Version]'7.0.0')
```

GitHubCore.ps1
```diff
@@ -315,7 +315,10 @@ function Invoke-GHRestMethod
             # Disable Progress Bar in function scope during Invoke-WebRequest
             $ProgressPreference = 'SilentlyContinue'
 
-            [Net.ServicePointManager]::SecurityProtocol=[Net.SecurityProtocolType]::Tls12
+            if ($PSVersionTable.PSVersion -lt [version]"7.0.0")
+            {
+                [Net.ServicePointManager]::SecurityProtocol=[Net.SecurityProtocolType]::Tls12
```
HowardWolosky: Fault in the origin code for not properly spacing this.

```suggestion
                [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```

HowardWolosky (dismissed):
I'm fine with this change I think, but when trying out PS Core 7.4.6 myself, I'm finding that TLS 1.2 is what is default, not 1.3.  Am I missing something?

GitHubCore.ps1
variableresistor: Likely it's getting set in some other module. According to Microsoft's documentation [SecurityProtocolType Enum](https://learn.microsoft.com/en-us/dotnet/api/system.net.securityprotocoltype), the default is SystemDefault. Specifically in the description, it says:
"Allows the operating system to choose the best protocol to use, and to block protocols that are not secure. Unless your app has a specific reason not to, you should use this value."

variableresistor: We COULD add an additional check:
```powershell
[Net.ServicePointManager]::SecurityProtocol -lt [System.Net.SecurityProtocolType]::Tls12
# So the full check would be:
if ($PSVersionTable.PSVersion -lt [version]'7.0.0' -and [Net.ServicePointManager]::SecurityProtocol -lt [System.Net.SecurityProtocolType]::Tls12)
{
    [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
}
```
What do you think?
EDIT: Changed double to single quotes

HowardWolosky: > We COULD add an additional check:
> 
> ```powershell
> [Net.ServicePointManager]::SecurityProtocol -lt [System.Net.SecurityProtocolType]::Tls12
> # So the full check would be:
> if ($PSVersionTable.PSVersion -lt [version]'7.0.0' -and [Net.ServicePointManager]::SecurityProtocol -lt [System.Net.SecurityProtocolType]::Tls12)
> {
>     [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
> }
> ```
> 
> What do you think? EDIT: Changed double to single quotes

I don't think that will quite work.  The problem I see there is that it will return `false` when it's set to `SystemDefault` (since its enum value is `0`), which then would ultimately have us back in the original state (from before this change) of setting the value to TLS 1.2.

The confusion that I have is the documentation (thanks for the link) indicates that `SystemDefault` is intended to allow the system to choose the best protocol to use, but I was previously forced to explicitly set it to TLS 1.2 because the system _wasn't_ making that choice, so I'm not sure how much confidence to put into that `SystemDefault` value.

variableresistor: How about this?
```powershell
if ($PSVersionTable.PSVersion -lt [version]'7.0.0' -and [Net.ServicePointManager]::SecurityProtocol -lt [System.Net.SecurityProtocolType]::Tls12)
{
  # Must do the comparison this way because of the Tls13 enumeration doesn't exist, then it will blow up
  if ( ([System.Net.SecurityProtocolType] | Get-Member -MemberType Property -Static).Name -contains 'Tls13')
  {
    [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls13
  }
  else
  {
    # Revert to the old behavior
    [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
  }
}
[Net.ServicePointManager]::SecurityProtocol
```
Worked on Windows PowerShell for me and
Would it slow things down too much? I figure in-memory comparisons are going to be far faster than disk or network operations. Added comments for us to refer to later.
EDIT: It's a bit slow, even when I try using native .NET methods :-(:
```powershell
Measure-Command { 'Tls13' -in ([System.Net.SecurityProtocolType] | Get-Member -MemberType Property -Static).Name }
Measure-Command { 'Tls13' -in [System.Net.SecurityProtocolType].GetEnumNames() }
```
```
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 8
Ticks             : 88331
TotalDays         : 1.02234953703704E-07
TotalHours        : 2.45363888888889E-06
TotalMinutes      : 0.000147218333333333
TotalSeconds      : 0.0088331
TotalMilliseconds : 8.8331

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 4
Ticks             : 44498
TotalDays         : 5.15023148148148E-08
TotalHours        : 1.23605555555556E-06
TotalMinutes      : 7.41633333333333E-05
TotalSeconds      : 0.0044498
TotalMilliseconds : 4.4498
```

variableresistor: For reference, this very simple command:
```powershell
Measure-Command { 'Tls13' | Out-Null }
```
Results:
```
Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 3
Ticks             : 30426
TotalDays         : 3.52152777777778E-08
TotalHours        : 8.45166666666667E-07
TotalMinutes      : 5.071E-05
TotalSeconds      : 0.0030426
TotalMilliseconds : 3.0426
```

HowardWolosky: I like the idea, and I think we can improve it.  While this answer is going to vary per system that the module is deployed to, it's not going to change during an individual session, so we don't need to check _every single_ time that we're going to do a network call.

We should capture the result to a static (`$script:PreferredSecurityProtocolType`) variable _once_ at the start of the session, and then just use that value going forward.  That being said, I'm not quite following your logic.  It's possible I'm missing something though.

If the user is on `< 7.0.0`, then we don't need to check their protocol version...we should just continue forcing TLS 1.2.
I _think_ what you were trying to do was this (the key being using `-ge` instead of `-lt`):

(this would likely go at the bottom of GitHubCore.ps1)
```powershell
$script:PreferredSecurityProtocolType = [Net.SecurityProtocolType]::Tls12
if (($PSVersionTable.PSVersion -ge [version]'7.0.0') -and
    (([System.Net.SecurityProtocolType] | Get-Member -MemberType Property -Static).Name -contains 'Tls13')
  {
    $script:PreferredSecurityProtocolType = [Net.SecurityProtocolType]::Tls13
  }
}
```

variableresistor: Again, I appreciate your time and effort on this change. I know it must be a lot of work maintaining this repository. So my team appreciates the work that's been done on this module.
This comparison from your code:
```powershell
$PSVersionTable.PSVersion -ge [Version]'7.0.0'
```
Is checking to see if we're running on PowerShell 7 and if we are, to set the protocol. In mine, what I want to to only tinker with the protocol if the user is running on Windows PowerShell.
The justification being that .NET 3.1 and higher already supports TLS 1.3 https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-3-0#security (at least on Linux) and PowerShell 7 doesn't even listen to this setting [Has there been a change in SSL/TLS backends with PowerShell 7 ?](https://github.com/PowerShell/PowerShell/discussions/16333#discussioncomment-1562157)
To also be on the conservative side, we could also check to make sure they're not running on on-prem deployment of GitHub:
```powershell
(Get-GitHubConfiguration -Name ApiHostName) -eq 'github.com'
```
I'm open to your thoughts on this.

variableresistor: It's tough because Microsoft actually recommends not fiddling with the protocol at all, as the case in our code, for when TLS 1.2 is inevitably exploited in the future [Transport Layer Security (TLS) best practices with the .NET Framework](https://learn.microsoft.com/en-us/dotnet/framework/network-programming/tls)

HowardWolosky: Ok, this is actually a bit more complicated than I had originally understood.  I appreciate the additional context.  I misread #443 and misunderstood your PR title, and had thought you were trying to have PS Core users take advantage of TLS 1.3 (I think the PR title and description could likely use a bit more flushing out for intent).

From a [little reading here](https://stackoverflow.com/questions/44870381/what-version-of-tls-is-used-in-a-net-application-installed-on-different-operati), it looks like .NET 4.7 is when it started to use the system default, but [.NET 4.6.2 is the first time that TLS 1.3 got introduced](https://learn.microsoft.com/en-us/dotnet/framework/network-programming/tls) (for Windows 11).

So, I believe the rough logic desired is as follows:

* Use System Default for PowerShell Core (7+)
* Use System Default for Windows PowerShell (< 7) if .NET version >= 4.7 *else* set to TLS 1.3 (if supported) *else* set to TLS 1.2.

Checking the .NET version [appears to be complicated](https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#the-environmentversion-property) appears to require a regkey check.

So, mapping that logic to code:

```powershell
# See https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed#minimum-version
$installedDotNetVersion = (Get-ItemPropertyValue -LiteralPath 'HKLM:SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full' -Name Release)
$dotNet47 = 460798

$script:PreferredSecurityProtocolType = [Net.SecurityProtocolType]::Tls12
if (($PSVersionTable.PSVersion -ge [version]'7.0.0') -or ($installedDotNetVersion -ge $dotNet47))
{
  # .NET 4.7 (460798) is the first .NET version to respect the system default.
  $script:PreferredSecurityProtocolType = [Net.SecurityProtocolType]::SystemDefault
}
elseif (([System.Net.SecurityProtocolType] | Get-Member -MemberType Property -Static).Name -contains 'Tls13')
{
  $script:PreferredSecurityProtocolType = [Net.SecurityProtocolType]::Tls13
}
```

Given this increased logic, my suggestion:

1. Wrap that logic inside of a new function (something like `Get-PreferredSecurityProtocolType`) inside of Helpers.ps1 (but have it return back a value, not modify a static.
2. Add a new read-only script variable to the top of GitHubCore.ps1 (below `ValidBodyContainingRequestMethods`) and have it's value come by calling that new helper method
3. Update `Invoke-GHRestMethod` to now use that script variable instead.

Thoughts?

variableresistor: Sounds like a plan Howard! Are we concerned about *nix compatibility? If so, checking for the PowerShell version first would short-circuit the if condition before PowerShell tries to check the registry that doesn't exist in Linux:

 

```powershell
$dotNet47 = 460798
$script:PreferredSecurityProtocolType = [Net.SecurityProtocolType]::Tls12

if (($PSVersionTable.PSVersion -ge [version]'7.0.0') -or ((Get-ItemPropertyValue -LiteralPath 'HKLM:SOFTWARE\Microsoft\NET Framework Setup\NDP\v4\Full' -Name Release) -ge $dotNet47))
{
  # .NET 4.7 (460798) is the first .NET version to respect the system default.
  $script:PreferredSecurityProtocolType = [Net.SecurityProtocolType]::SystemDefault
}
elseif (([System.Net.SecurityProtocolType] | Get-Member -MemberType Property -Static).Name -contains 'Tls13')
{
  $script:PreferredSecurityProtocolType = [Net.SecurityProtocolType]::Tls13
}
```

HowardWolosky: Good callout.  Yes -- this module should be compatible across all PS Core supported platforms.
Your modified update generally looks good...to keep line length under 80 though, you'll likely need to move the `((Get-ItemPropertyValue ...` to the subsequent line, indented.  Please also retain the comment with the link to the MSDN documentation as well, and remember to update this to be a helper function that ultimately returns a value instead of directly setting the static...have the static set with the resulting value from the function call.

variableresistor: Verified that this works in regular Windows PowerShell by running:
```powershell
# This is returned from Get-PreferredSecurityProtocolType
[Net.ServicePointManager]::SecurityProtocol

Invoke-RestMethod https://github.com
Get-PreferredSecurityProtocolType
# Returns SystemDefault
```
No error. 

I fixed an error about a missing repo variable that is due to strict mode being set as well. There were a few errors other I got, but were unrelated to my change:
```
[-] Should have the expected type and additional properties 1.29s (1.28s|10ms)
  Expected strings to be the same, but they were different.
  Expected length: 17
  Actual length:   52
  Strings differ at index 0.
  Expected: 'GitHub.GistCommit'
  But was:  'Selected.System.Management.Automation.PSCustomObject'
               ^
  at $commit.PSObject.TypeNames[0] | Should -Be 'GitHub.GistCommit', 
C:\Repos\PowerShellForGitHub\Tests\GitHubGists.tests.ps1:163
  at <ScriptBlock>, C:\Repos\PowerShellForGitHub\Tests\GitHubGists.tests.ps1:163

[-] Should have fewer results with using the since parameter 470ms (468ms|2ms)
  Expected the actual value to be greater than 0, but got $null.
  at $sinceGists.Count | Should -BeGreaterThan 0, C:\Repos\PowerShellForGitHub\Tests\GitHubGists.tests.ps1:254
  at <ScriptBlock>, C:\Repos\PowerShellForGitHub\Tests\GitHubGists.tests.ps1:254

  [-] Should throw the correct exception 580ms (575ms|5ms)
   Expected like wildcard '*The remote server returned an error: (401) Unauthorized*' to match 'The remote server returned an error: (403) Forbidden.
   API rate limit exceeded for 172.117.119.161. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.) | https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting
   RequestId: C0D8:2809CF:B467BE:B96E51:67912352', but it did not match.
   at $Error[0].Exception.Message | Should -BeLike $exceptionMessage1, C:\Repos\PowerShellForGitHub\Tests\GitHubGraphQl.Tests.ps1:156
   at <ScriptBlock>, C:\Repos\PowerShellForGitHub\Tests\GitHubGraphQl.Tests.ps1:156

[-] Should throw the correct exception 684ms (680ms|4ms)
  Expected like wildcard '*Parse error on "InvalidQuery"*' to match 'GraphQl Error: Expected one of SCHEMA, SCALAR, TYPE, ENUM, INPUT, UNION, INTERFACE, actual: IDENTIFIER ("InvalidQuery") at [1, 1]
  RequestId: C20C:3D330C:B18DD3:B6944B:67912353', but it did not match.
  at $Error[0].Exception.Message | Should -BeLike "*Parse error on ""$invalidQuery""*", C:\Repos\PowerShellForGitHub\Tests\GitHubGraphQl.Tests.ps1:201
  at <ScriptBlock>, C:\Repos\PowerShellForGitHub\Tests\GitHubGraphQl.Tests.ps1:201
```

variableresistor: Also on the topic of Helpers.ps1, can I just get rid of the Ensure-Directory function? It's not referenced anywhere and doesn't use an approved verb.

